PR #19, opened 2025-05-07
chore: Add Jolt

kevaundray:


kevaundray:
This is partially implemented in Ere, however there are some issues with Jolt proving

PR #20, opened 2025-05-07
chore: Add Pico

kevaundray:


kevaundray:
Keeping this open for the precompiles, but its implemented in Ere

PR #27, opened 2025-05-17
chore: Migrate zkevm-sp1 to Ere

kevaundray:

PR #29, opened 2025-05-18
chore: Migrate to Ere

kevaundray:


crates/ere-hosts/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+    // Iterate through test corpus and generate reports
+    for bw in &corpuses {
+        println!(" {} ({} blocks)", bw.name, bw.blocks_and_witnesses.len());
+
+        let mut reports = Vec::new();
+        for ci in &bw.blocks_and_witnesses {
+            let mut stdin = Input::new();
+            stdin.write(ci)?;
+            stdin.write(&bw.network)?;
+
```
kevaundray, line 50: Genesis contains fields that are marked with skip_serializing_if, so we either need to bring back ForkSpec or use a different serialization strategy

kevaundray: Reverted this and went back to forkspec for now

crates/ere-hosts/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+    // Compile program and create proving/verification keys
+    let program = C::compile(&PathBuf::from(guest_dir))?;
+    let zkvm_instance = V::new(program);
+    let zkvm_ref = Arc::new(&zkvm_instance);
+
+    let corpuses = generate_stateless_witness::generate();
+
+    // Iterate through test corpus and generate reports
+    // TODO: note that when proving, processing these in parallel will likely skew the results
```
kevaundray, line 43: We can probably switch to serial processing and get a powerful server to run these on

crates/ere-hosts/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+    benchmark::<ere_sp1::RV32_IM_SUCCINCT_ZKVM_ELF, ere_sp1::EreSP1>(
+        "sp1",
+        concat!(env!("CARGO_WORKSPACE_DIR"), "ere-guests/sp1"),
+    )?;
+
+    // TODO: Add more backends
+    Ok(())
+}
+
+// TODO: Eventually move this into benchmark_runner
```
kevaundray, line 25: This would be the equivalent of the function that Ignacio made

crates/ere-hosts/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+
+/// Main entry point for the host benchmarker
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    dotenv::dotenv().ok();
+
+    // SP1 zkVM
+    benchmark::<ere_sp1::RV32_IM_SUCCINCT_ZKVM_ELF, ere_sp1::EreSP1>(
+        "sp1",
+        concat!(env!("CARGO_WORKSPACE_DIR"), "ere-guests/sp1"),
+    )?;
```
kevaundray, line 19: - ere_sp1::RV32_IM_SUCCINCT_ZKVM_ELF : This is the thing that compiles the rust code
- ere_sp1::EreSP1 : This is the thing that executes, proves and verifies
- "sp1" : This is the name of the zkvm
-  concat!(env!("CARGO_WORKSPACE_DIR"), "ere-guests/sp1") : This is where the guest program is located that we want to compile

kevaundray: We can probably make this binary nicer by allowing people to do `cargo run --release --zkvm sp1 risc0` but not high priority and not blocking for this

kevaundray:
supercedes #27

kevaundray:
We can also deprecate WorkloadMetrics since ere has ExecutionReport which has the same information

PR #40, opened 2025-05-24
feat: Refine benchmark runner API

kevaundray:
Main change is that we can specify execute or proving


kevaundray:
We can make a followup PR to make some of the fields in WokloadMetrics optional;  proving only uses name and proving_time. While execution uses the other fields

kevaundray:
Followup PR is #41

PR #50, opened 2025-05-24
chore: metrics skunkworks

kevaundray:


jsign:
We recently added some python scripts with this intention, so I think we can close this.
We can keep iterating on them if we want to tune them further.

PR #59, opened 2025-05-26
chore: add analytics

kevaundray:


jsign:
We have the https://zkevm-analysis.vercel.app/ app. 
I'm not sure if makes sense to upload this to the repo really, sounds more like a separate application, and not sure will be maintained long term.

In any case, I think we can still close this PR @kevaundray? (if that's the case, feel free to close it).

PR #62, opened 2025-05-28
chore: r0

kevaundray:
To run:

```
- cd crates/ere-hosts
- Change the feature flag of `ere-risczero` to be metal or cuda, depending on hardware. Its currently set to metal
- RUSTFLAGS="-C target-cpu=native" cargo run --release  -- --zkvm risc0 --resource gpu --action prove
```

### DIFF
--- a/crates/ere-hosts/Cargo.toml
+++ b/crates/ere-hosts/Cargo.toml
@@ -8,10 +8,10 @@ license.workspace = true
 [dependencies]
 benchmark-runner.workspace = true
 ere-sp1.workspace = true
-ere-risczero.workspace = true
+ere-risczero = { workspace = true, features = ["metal"] }
 # ere-pico.workspace = true
 zkvm-interface.workspace = true
-ere-openvm.workspace = true
+#ere-openvm.workspace = true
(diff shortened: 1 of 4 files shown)


crates/ere-hosts/src/main.rs
```diff
@@ -122,13 +122,13 @@ fn new_risczero_zkvm(
     Ok(EreRisc0::new(program, prover_resource))
 }
 
-fn new_openvm_zkvm(
```
kevaundray, line 125: You can ignore openVM being commented out. openvm and r0 rely on the same low level c syscall functions so we can't  compile them together

crates/ere-hosts/Cargo.toml
```diff
@@ -8,10 +8,10 @@ license.workspace = true
 [dependencies]
 benchmark-runner.workspace = true
 ere-sp1.workspace = true
-ere-risczero.workspace = true
+ere-risczero = { workspace = true, features = ["metal"] }
```
kevaundray, line 11: We can change "metal" to "cuda" to make it work with cuda

PR #66, opened 2025-06-04
Add `ere-guests/zisk`

han0110:
- Order cargo alias alphabetically so it is easier to track.
- Add `ere-guests/zisk`, but not sure if it's has the necessary output for the benchmark to collect.
- Add `precompile-patches/zisk.toml`, but in org [0xPolygonHermez](https://github.com/0xPolygonHermez) there is only `tiny-keccak` patched repo, and an outdated patched [reth](https://github.com/0xPolygonHermez/zisk-patch-reth), not sure we should do the rest patches for them (e.g. `k256`). The supported precompiled could be found [here](https://0xpolygonhermez.github.io/zisk/getting_started/precompiles.html#available-precompiles-in-zisk).
- Add `zkVM::Zisk` in `ere_hosts`.

Todo: 
- [x] Merge https://github.com/eth-applied-research-group/ere/pull/31
- [x] Revert the `ere` dependency changes

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,7 +2,7 @@
 CARGO_WORKSPACE_DIR = { value = "", relative = true }
 
 [alias]
+pico = ["run", "-p", "xtask", "--", "pico", "--"]
+risc0 = ["run", "-p", "xtask", "--", "risc0", "--"]
 sp1 = ["run", "-p", "xtask", "--", "sp1", "--"]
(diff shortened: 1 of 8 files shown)


kevaundray:
> not sure we should do the rest patches for them (e.g. k256). 

imo I don't think its necessary right now

kevaundray (approved):

PR #70, opened 2025-06-05
risc0 getrandom and ere update

jsign:
This PR:
- Updates `ere` and pins to a specific commit (latest now), so we don't have a non-well defined dependency target in case of breaking changes happening.
- Fixes risc0 guest program to allow for `getrandom` feature.

The last bullet was discovered when I tried to run mainnet blocks with risc0 and got this panic:
```
thread 'main' panicked at /home/ignacio/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/risc0-zkvm-platform-2.0.2/src/getrandom.rs:63:5:


WARNING: `getrandom()` called from guest.
=========================================

This panic was generated from the guest program calling `getrandom()`.
Using randomness in the zkVM requires careful attention to your
application’s security requirements. Therefore, the default behavior is
for the zkVM to generate this panic message when the guest calls
`getrandom()`. If you wrote this guest program, it is crucial that you
understand the implications of using randomness in your guest code and
make changes to fit your needs.


The zkVM supports providing random data from the host in response to
`getrandom()`. When the randomness is being used to protect private data, this
is a good option. However, it is not appropriate for all use cases. Consider a
situation when random data is needed to ensure the honesty of the prover (e.g.
to flip a coin for a game between the prover and verifier). In this scenario,
host provided randomness is not suitable because the prover may alter the source
of randomness. For such use cases, great care must be taken to provide a source
of randomness that the prover cannot manipulate or predict. Host provided
randomness can be enabled with the `getrandom` feature flag on the `risc0-zkvm`
crate used for the guest.


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Benchmark CRASHED for mainnet_block_22639757: Error: Guest panicked: 

WARNING: `getrandom()` called from guest.
=========================================

This panic was generated from the guest program calling `getrandom()`.
Using randomness in the zkVM requires careful attention to your
application’s security requirements. Therefore, the default behavior is
for the zkVM to generate this panic message when the guest calls
`getrandom()`. If you wrote this guest program, it is crucial that you
understand the implications of using randomness in your guest code and
make changes to fit your needs.


The zkVM supports providing random data from the host in response to
`getrandom()`. When the randomness is being used to protect private data, this
is a good option. However, it is not appropriate for all use cases. Consider a
situation when random data is needed to ensure the honesty of the prover (e.g.
to flip a coin for a game between the prover and verifier). In this scenario,
host provided randomness is not suitable because the prover may alter the source
of randomness. For such use cases, great care must be taken to provide a source
of randomness that the prover cannot manipulate or predict. Host provided
randomness can be enabled with the `getrandom` feature flag on the `risc0-zkvm`
crate used for the guest.
```

The message is quite clear, but if you are interested in what is happening under the hood here's what I found:
- The [`getrandom` crate](https://docs.rs/getrandom/0.3.3/getrandom) has a [special feature `custom`](https://docs.rs/getrandom/0.3.3/getrandom/#opt-in-backends) to allow arbitrary implementations.
- The build package of risc0 [uses this compilation flag](https://github.com/risc0/risc0/blob/main/risc0/build/src/lib.rs#L493) to enable the feature. So it has to implement that [custom backend](https://docs.rs/getrandom/0.3.3/getrandom/#custom-backend).
- If you enable the `getrandom` feature of `zkvm-platform`, you change that [custom backend implementation](https://github.com/risc0/risc0/blob/main/risc0/zkvm/platform/src/getrandom.rs#L17-L91). As you can see there, since isn't enabled by default that last link has the panic message that I showed before. If you enable it, then puts an implementation that calls the `sys_rand` syscall.

All makes sense as a way to control if the `sys_rand` syscall is allowed. Was quite cool to discover that the `getrandom` craate (which is a public crate) allows to set custom backends that could wire all this.

From our repo side just enabling `getrandom` is the fix and now we can execute mainnet blocks. I would like to eventually dig why for running test cases this wasn't happening, but probably there's a reasonable explanation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,11 +111,11 @@ witness-generator = { path = "crates/witness-generator" }
 zkevm-metrics = { path = "crates/metrics" }
 benchmark-runner = { path = "crates/benchmark-runner" }
 
-zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", package = "zkvm-interface" }
-ere-sp1 = { git = "https://github.com/eth-applied-research-group/ere", package = "ere-sp1" }
-ere-risczero = { git = "https://github.com/eth-applied-research-group/ere", package = "ere-risczero" }
-ere-pico = { git = "https://github.com/eth-applied-research-group/ere", package = "ere-pico" }
-ere-openvm = { git = "https://github.com/eth-applied-research-group/ere", package = "ere-openvm" }
+zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", rev = "5f4fe5e7a83c969cab160b10644330ab771ee94c", package = "zkvm-interface" }
+ere-sp1 = { git = "https://github.com/eth-applied-research-group/ere", rev = "5f4fe5e7a83c969cab160b10644330ab771ee94c", package = "ere-sp1" }
+ere-risczero = { git = "https://github.com/eth-applied-research-group/ere", rev = "5f4fe5e7a83c969cab160b10644330ab771ee94c", package = "ere-risczero" }
+ere-pico = { git = "https://github.com/eth-applied-research-group/ere", rev = "5f4fe5e7a83c969cab160b10644330ab771ee94c", package = "ere-pico" }
+ere-openvm = { git = "https://github.com/eth-applied-research-group/ere", rev = "5f4fe5e7a83c969cab160b10644330ab771ee94c", package = "ere-openvm" }
 
 # branch is kw/zkevm-benchmark-workload-repo
 # NOTE: We are using a branch of a branch that has not yet been merged into master.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1700,8 +1700,7 @@ dependencies = [
 [[package]]
 name = "blst"
 version = "0.3.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -111,11 +111,11 @@ witness-generator = { path = "crates/witness-generator" }
 zkevm-metrics = { path = "crates/metrics" }
 benchmark-runner = { path = "crates/benchmark-runner" }
 
-zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", package = "zkvm-interface" }
-ere-sp1 = { git = "https://github.com/eth-applied-research-group/ere", package = "ere-sp1" }
-ere-risczero = { git = "https://github.com/eth-applied-research-group/ere", package = "ere-risczero" }
-ere-pico = { git = "https://github.com/eth-applied-research-group/ere", package = "ere-pico" }
-ere-openvm = { git = "https://github.com/eth-applied-research-group/ere", package = "ere-openvm" }
+zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", rev = "5f4fe5e7a83c969cab160b10644330ab771ee94c", package = "zkvm-interface" }
```
jsign, line 114: I guess at some point we can start creating tags.

kevaundray (approved):
LGTM!

Cargo.toml
kevaundray: Yep, and or publishing to crates.io

PR #71, opened 2025-06-05
Generate witness once before running benchmarks

jsign:
I was running benchmarks for two zkVMs for mainnet blocks and I realized the way things were wasn't ideal.

The witness generation call was done on each benchmark zkVM run, which for mainnet blocks means each will pull different blocks since we pul the last N blocks.

This PR changes the logic to generate the witness once, and use the same corpus on each zkVM. This also avoid generating witnesses repeatedly which would speed up benchmarks a bit too. And also, is a small step in the direction of untangling witness generation from benchmark runs (e.g. new benchmark running is not async again).

### DIFF
--- a/crates/benchmark-runner/src/lib.rs
+++ b/crates/benchmark-runner/src/lib.rs
@@ -51,18 +51,16 @@ pub enum Action {
     Execute,
 }
 
-pub async fn run_benchmark_ere<V>(
+pub fn run_benchmark_ere<V>(
     host_name: &str,
     zkvm_instance: V,
     action: Action,
-    wg: &Box<dyn WitnessGenerator>,
-) -> Result<()>
-where
+    corpuses: &[BlocksAndWitnesses],
+) where
     V: zkVM + Sync,
 {
     println!("Benchmarking `{}`…", host_name);
     let zkvm_ref = Arc::new(&zkvm_instance);
-    let corpuses = wg.generate().await?;
 
     match action {
         Action::Execute => {
@@ -78,12 +76,10 @@ where
             });
         }
     }
-
-    Ok(())
 }
 
 fn process_corpus_with_crash_handling<V>(
-    bw: BlocksAndWitnesses,
+    bw: &BlocksAndWitnesses,
     zkvm_ref: Arc<&V>,
     action: &Action,
     host_name: &str,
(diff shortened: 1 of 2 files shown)


crates/benchmark-runner/src/lib.rs
```diff
@@ -175,12 +171,12 @@ where
         None => panic!("unexpected test with no blocks {}", &bw.name),
     };
 
-    bw.blocks_and_witnesses = vec![last_block_with_witness];
+    let blocks_and_witnesses = vec![last_block_with_witness];
```
jsign, line 174: Not sure if this vec with a single entry was done with a particular purpose originally, but I'm simply honoring the original style. If there's no good purpose of longer chain runs, we can refactor a bit and run only the latest blocks.

For now leaving it this way -- food for thought.

jsign (commented):
@kevaundray, not touching the docs to avoid a conflict with your PR. 

Also, no rush to merge this if you want to merge yours first. I can rebase after.

crates/benchmark-runner/src/lib.rs
kevaundray: Ah this was basically because with the execution spec bnchmarks, we only want the last block

kevaundray (approved):

crates/ere-hosts/src/main.rs
```diff
@@ -142,37 +142,39 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
     // Set to true once a zkvm has ran
     let mut ran_any = false;
 
     #[cfg(feature = "sp1")]
     {
         run_cargo_patch_command("sp1")?;
         let sp1_zkvm = new_sp1_zkvm(resource)?;
-        run_benchmark_ere("sp1", sp1_zkvm, action, &block_witness_gen).await?;
+        run_benchmark_ere("sp1", sp1_zkvm, action, &corpuses).await?;
```
jsign, line 154: @kevaundray, ah I think this merge with the `master` branch is wrong.
It shouldn't have the `.await` since isn't async anymore.

I think this wasn't catch by the CI since prob we need to tune to enable all features or at least 1?

kevaundray: Ah sorry, yeah you're right

PR #73, opened 2025-06-06
Fix `ere-hosts` with `feature = zisk`

han0110:
- Apply the change of #72 to `ere-hosts` with `feature = zisk`
- Add CI checks for each zkvm for `ere-hosts` (it increases x secs to run the CI but I think it's worth doing)

### DIFF
--- a/.github/workflows/rust_checks.yml
+++ b/.github/workflows/rust_checks.yml
@@ -15,17 +15,17 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-      
+
       - name: Download and Extract Fixtures
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/rust_checks.yml
```diff
@@ -38,12 +38,50 @@ jobs:
...
+  check-ere-host:
+    name: Checks crate ere-hosts
+    strategy:
+      matrix:
+        include:
+          - zkvm: sp1
+          - zkvm: risc0
+          - zkvm: openvm
+          - zkvm: pico
+          - zkvm: zisk
```
jsign, line 57: Nice!

PR #74, opened 2025-06-12
feat(openvm): add precompile patches

jonathanpwang:
This PR adds the necessary guest setup and patches to trigger all OpenVM custom instructions and patches for ETH block validation. Updates to `openvm v1.4.0`.

We currently patch `revm` to handle pre-compiles, but in a subsequent PR we will switch to using the new `CryptoProvider` trait to be able to directly use upstream `revm`.

I was not able to include the openvm `ruint` patch because it requires version to be fixed to `v1.14.0` but some other risc0 packages seem to require v1.15^. From previous profiling, the ruint patch is not one of the most significant patches, so it's okay to skip it for now (although performance is still best with it).

We note that currently the `stateless_validation` function has a significant performance degradation because the conversion from execution witness to MPT in the guest program is very unoptimized, compared to our current deserialization approach:
**Flamegraphs to be updated** but we believe the performance degradation remains
- deserialization approach (faster) [benchmark](https://github.com/axiom-crypto/openvm-reth-benchmark/blob/gh-pages/benchmarks-dispatch/refs/heads/chore/foldhash/reth-a088dfe491b30a6d7925f9b3d5a14777e3f88fc4-b1fcf810ab3987472c37bfc22b8d27775c99152b4a7098a236fe0e00e750faf0.md): 

The flamegraph is weighted by number of AIR trace cells, not instructions, per call stack. [source](https://github.com/axiom-crypto/openvm-reth-benchmark/blob/gh-pages/benchmarks-dispatch/refs/heads/main/reth-357041628ad40f5bc3a02b0835e7bc72e716b069-f06666e653e1195fec07add609aac2114f20b1e7865e542fd3a8ddd137d1cd2b.md)

- `stateless_validation` approach (slower) [benchmark]() run from [this branch](https://github.com/axiom-crypto/openvm-reth-benchmark/pull/423):

Observe the large portion spent on `StatelessTrie::new()`. [source](https://github.com/axiom-crypto/openvm-reth-benchmark/blob/gh-pages/benchmarks-dispatch/refs/heads/val/reth-stateless/reth-631ea41abb37e5ad1a7973fb40bd16055535d491-02ca23cf16f16dfe48b1ef00b28bf17f91f0042dc5ed0beda9fdc22150344c98.md)


jonathanpwang:
@jsign I've updated to OpenVM v1.4.0.

I wasn't able to test it directly in the repo yet because the readme instructions:
```
cd crates/witness-generator-cli
cargo run --release -- tests --include Prague --include cold
```
gave me error
```
Error: Failed to build EEST generator

Caused by:
    0: Failed to execute download script
    1: No such file or directory (os error 2)
```

By the way, with OpenVM v1.4.0, the CLI and SDK both come with GPU support if feature `cuda` is enabled: https://docs.openvm.dev/book/getting-started/install#installation-with-nvidia-gpu-support 
I haven't looked into how this is handled in `ere-host` yet, so just mentioning it's an option. cc @han0110

jsign:
@jonathanpwang, thanks very much!

Ah, regarding `witness-generator-cli` we should fix the docs -- should be ran at the workspace level, not crate level. Standing in root of the repo, running `RAYON_NUM_THREADS=10 RUST_LOG=info cargo run --release -p witness-generator-cli -- tests --include 10M- --include Prague` will generate all 10M gas tests (as an example).

Reg OpenVM v1.4.0, we just merged the support for `ere` today in https://github.com/eth-act/ere/pull/113 :) We have to update the `ere` dep in this repo, and should be good to go.

I'm OOO until Monday, so I can do this as soon as I come back.

jonathanpwang:
@han0110 yes I noted it in the PR description. I tried to patch to 1.16 but there was some change that requires `const fn` and it was not the most straightforward. I believe it doesn't affect performance too much because it's all in `revm_interpreter::arithmetic` and even there most of the instructions are associated with handling the stack

jsign:
The CI job failing is expected since this is an external branch PR, so the GH repo secrets are guarded. Thanks very muc hagain @jonathanpwang.

jsign:
@jonathanpwang, I'm updating our fork to latest Reth in #182 -- but found a problem with OpenVM compiler, more see comment https://github.com/eth-act/zkevm-benchmark-workload/pull/182/files#r2337888495.

PR #75, opened 2025-06-13
feat(ere-hosts): `--block` parameter for fetching specific blocks

shekhirin:
```console
cargo run --bin ere-hosts --features sp1 -- rpc --block 22691734 --rpc-url https://reth-ethereum.ithaca.xyz/rpc
```


jsign (approved):
LGTM! 

Originally, I didn't add this option since I noticed reth has somewhat degraded performance the longer I go back in history.

In any case, I think it is good to include the feature. 

Thanks!

PR #82, opened 2025-06-22
Include host hardware spec, block used gas, proof size and have more uniform crash handling in outputs

jsign:
This PR does the following changes:

- Automatically detects CPU, RAM, and GPUs in the host machine and saves it in a `hardware.json` file at the root of `zkevm-metrics` output folder.
- Adds an extra field in each benchmark run output file named `gas_used` with the amount of gas used in the block. This is useful for later analytics (e.g., having a proving time of 1min in a 10M block is not the same as a 30M block, especially for mainnet block benchmark runs). Also, I'm interested in presenting "proving opcode X has X gas/s proving throughput" which can in trying to make sense of potential repricings.
- Avoid using a separate `crashed` folder for crashed runs. Every run now will have the expected output json file, and will signal `success` or `crashed`. (More about this in example outputs below)
- The output file accepts `execution` and `proving` information, allowing us to support running benchmark both in execution and proving mode. This can be useful to have both details about total/region cycle information in a single file, plus providing information so that the analytics can have everything in a single file. (Will be more evident in the example below).
- Adds the field `proof_size` in proving action mode.

Let’s look at some examples that will make all the above clearer.

At the root of the output folder, a new `hardware.json` is generated like:
```json
{
  "cpu_model": "AMD EPYC 7R13 Processor",
  "total_ram_gib": 248,
  "gpus": [
    {
      "model": "NVIDIA L40S"
    }
  ]
}
```
```
zkevm-metrics
├── hardware.json
└── sp1
    └── tests
        └── zkevm
        ...
```

This is written as a separate file compared to earlier drafts, which always included this (repeated) information in each output file. The future analytics tool can read this file once to contextualize the runs on which hardware was run.

Regarding benchmark outputs, here’s a successful one for an execution run:

```json
[
  {
    "name": "tests/zkevm/test_worst_stateful_opcodes.py::test_worst_address_state_cold[fork_Prague-blockchain_test-absent_accounts_False-opcode_BALANCE]-2",
    "block_used_gas": 36000000,
    "execution": {
      "success": {
        "total_num_cycles": 699424217,
        "region_cycles": {
          "validation": 435721514,
          "decode-headers": 9720,
          "recover-signers": 51887,
          "verify-witness": 287496071,
          "post-state-compute": 525030,
          "read_input": 263698439,
          "block-execution": 144356703
        },
        "execution_duration": {
          "secs": 37,
          "nanos": 479416950
        }
      }
    }
  }
]
```

Note `gas_used`, and `success`.

Compare with a crashed execution output (as a normal output file, i.e., not in a special folder):

```json

[
  {
    "name": "tests/zkevm/test_worst_stateful_opcodes.py::test_worst_address_state_warm[fork_Prague-blockchain_test-absent_target_False-opcode_CALL]-1",
    "block_used_gas": 36000000,
    "execution": {
      "crashed": {
        "reason": "SP1 execution failed: execution failed with exit code 1"
      }
    }
  }
]
```

You can see `crashed` with the detailed message. The benefit is that the analyzer can know what happened all in a single file, without looking at different folders. 

Proving run example:

```json
[
  {
    "name": "rpc_block_22756114-22756114",
    "block_used_gas": 34698073,
    "proving": {
      "success": {
        "proof_size": 1477199,
        "proving_time_ms": 734473
      }
    }
  }
]
```

note the new `proof_size` field.

As mentioned before in the bullets, the format of this output is:

```rust
/// Represents a single benchmark run.
#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
pub struct BenchmarkRun {
    /// Name of the benchmark.
    pub name: String,
    /// Block used gas
    pub block_used_gas: u64,
    /// Execution metrics for the benchmark run.
    #[serde(skip_serializing_if = "Option::is_none")]
    pub execution: Option<ExecutionMetrics>,
    /// Proving metrics for the benchmark run.
    #[serde(skip_serializing_if = "Option::is_none")]
    pub proving: Option<ProvingMetrics>,
}
```

so both `execution` and `proving` are top-level fields. They are optional depending on which action is passed to the tool. This potentially allows for supporting both in the same output file. The CLI still only asks for one action kind — didn’t want to add this feature yet to avoid making the PR bigger, but we can support it in a future PR.

### DIFF
--- a/crates/benchmark-runner/src/lib.rs
+++ b/crates/benchmark-runner/src/lib.rs
@@ -68,6 +68,7 @@ where
 
     for ci in blocks_and_witnesses {
         let block_number = ci.block.number;
+        let block_used_gas = ci.block.gas_used;
         let mut stdin = Input::new();
         stdin.write(ci);
         stdin.write(bw.network);
@@ -109,6 +110,7 @@ where
         };
         reports.push(BenchmarkRun {
             name: format!("{}-{}", bw.name, block_number),
+            block_used_gas,
             hardware: hardware.clone(),
             actions_metrics: vec![workload_metrics],
         });
--- a/crates/benchmark-runner/src/lib.rs
+++ b/crates/benchmark-runner/src/lib.rs
@@ -193,9 +193,10 @@ where
                 }
             }
             Action::Prove => {
-                let (_, report) = zkvm_ref.prove(&stdin)?;
+                let (proof, report) = zkvm_ref.prove(&stdin)?;
                 WorkloadMetrics::Proving {
                     name: format!("{}-{}", bw.name, block_number),
+                    proof_size: proof.len(),
                     proving_time_ms: report.proving_time.as_millis(),
                 }
             }
--- a/crates/benchmark-runner/src/lib.rs
+++ b/crates/benchmark-runner/src/lib.rs
@@ -33,13 +33,13 @@ where
             // Use parallel iteration for execution
             corpuses
                 .into_par_iter()
-                .try_for_each(|bw| process_corpus(bw, zkvm_ref.clone(), &action, host_name))?;
+                .try_for_each(|bw| process_corpus(bw, zkvm_ref.clone(), action, host_name))?;
         }
         Action::Prove => {
             // Use sequential iteration for proving
             corpuses
                 .into_iter()
-                .try_for_each(|bw| process_corpus(bw, Arc::new(&*zkvm_ref), &action, host_name))?;
+                .try_for_each(|bw| process_corpus(bw, zkvm_ref.clone(), action, host_name))?;
         }
     }
     Ok(())
@@ -47,8 +47,8 @@ where
 
 fn process_corpus<V>(
     bw: &BlocksAndWitnesses,
-    zkvm_ref: Arc<&V>,
-    action: &Action,
+    zkvm_ref: Arc<V>,
+    action: Action,
     host_name: &str,
 ) -> anyhow::Result<()>
 where
@@ -56,10 +56,11 @@ where
 {
     // Take the last element, because benchmarks are setup in such a way that
     // We only want to benchmark the last block.
-    let last_block_with_witness = match bw.blocks_and_witnesses.last() {
-        Some(last_block) => last_block.clone(),
-        None => panic!("unexpected test with no blocks {}", &bw.name),
-    };
+    let last_block_with_witness = bw
+        .blocks_and_witnesses
+        .last()
+        .ok_or_else(|| anyhow::anyhow!("Unexpected test with no blocks: {}", &bw.name))?
+        .clone();
 
     let blocks_and_witnesses = vec![last_block_with_witness];
 
@@ -128,11 +129,9 @@ where
 }
 
 fn get_panic_msg(panic_info: Box<dyn Any + Send>) -> String {
-    if let Some(s) = panic_info.downcast_ref::<&str>() {
-        s.to_string()
-    } else if let Some(s) = panic_info.downcast_ref::<String>() {
-        s.clone()
-    } else {
-        "Unknown panic occurred".to_string()
-    }
+    panic_info
+        .downcast_ref::<&str>()
+        .map(|s| s.to_string())
+        .or_else(|| panic_info.downcast_ref::<String>().cloned())
+        .unwrap_or_else(|| "Unknown panic occurred".to_string())
 }
--- a/crates/benchmark-runner/src/lib.rs
+++ b/crates/benchmark-runner/src/lib.rs
@@ -16,7 +16,8 @@ pub fn run_benchmark_ere<V>(
     zkvm_instance: V,
     action: Action,
     corpuses: &[BlocksAndWitnesses],
-) where
+) -> anyhow::Result<()>
+where
     V: zkVM + Sync,
 {
     println!("Benchmarking `{}`…", host_name);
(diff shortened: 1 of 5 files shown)


crates/benchmark-runner/src/lib.rs
```diff
@@ -1,49 +1,9 @@
...
 use rayon::prelude::*;
-use std::{fs, panic, sync::Arc};
-use witness_generator::{witness_generator::WitnessGenerator, BlocksAndWitnesses};
-use zkevm_metrics::WorkloadMetrics;
+use std::{any::Any, panic, sync::Arc};
+use witness_generator::BlocksAndWitnesses;
+use zkevm_metrics::{BenchmarkRun, CrashInfo, ExecutionMetrics, HardwareInfo, ProvingMetrics};
 use zkvm_interface::{zkVM, Input};
 
-#[deprecated(note = "this function is being phased out, use run_benchmark_ere")]
```
jsign, line 8: Removed unused method.

crates/benchmark-runner/src/lib.rs
```diff
@@ -56,120 +16,51 @@ pub fn run_benchmark_ere<V>(
     zkvm_instance: V,
     action: Action,
     corpuses: &[BlocksAndWitnesses],
-) where
+) -> anyhow::Result<()>
+where
     V: zkVM + Sync,
 {
+    HardwareInfo::detect().to_path(format!(
```
jsign, line 23: Save the `hardware.json` file in the root `zkevm-metrics` folder.

I still we need some extra work to make the output folder of the benchmark be more encapsulated, but we can do those improvements in future PRs. For now keeping the same style as before.

crates/benchmark-runner/src/lib.rs
```diff
@@ -56,120 +16,51 @@ pub fn run_benchmark_ere<V>(
...
         Action::Prove => {
             // Use sequential iteration for proving
-            corpuses.into_iter().for_each(|bw| {
-                process_corpus_with_crash_handling(bw, Arc::new(&*zkvm_ref), &action, host_name);
-            });
-        }
-    }
-}
-
-fn process_corpus_with_crash_handling<V>(
```
jsign, line 81: This method is gone too, the crash handling is done inside each benchmark run.

crates/benchmark-runner/src/lib.rs
```diff
@@ -56,120 +16,51 @@ pub fn run_benchmark_ere<V>(
...
     }
+    Ok(())
 }
 
 fn process_corpus<V>(
     bw: &BlocksAndWitnesses,
-    zkvm_ref: Arc<&V>,
-    action: &Action,
+    zkvm_ref: Arc<V>,
+    action: Action,
```
jsign, line 51: Just some simplifications (e.g. `Action` is `Copy`).

crates/benchmark-runner/src/lib.rs
```diff
@@ -56,120 +16,51 @@ pub fn run_benchmark_ere<V>(
...
     // We only want to benchmark the last block.
-    let last_block_with_witness = match bw.blocks_and_witnesses.last() {
-        Some(last_block) => last_block.clone(),
-        None => panic!("unexpected test with no blocks {}", &bw.name),
-    };
+    let last_block_with_witness = bw
+        .blocks_and_witnesses
+        .last()
+        .ok_or_else(|| anyhow::anyhow!("Unexpected test with no blocks: {}", &bw.name))?
+        .clone();
```
jsign, line 63: A bit nicier and remove the panic.

crates/ere-hosts/src/main.rs
```diff
@@ -165,39 +165,39 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        run_benchmark_ere("sp1", sp1_zkvm, action, &corpuses)?;
         ran_any = true;
     }
 
     #[cfg(feature = "zisk")]
     {
         run_cargo_patch_command("zisk")?;
-        let zisk_zkvm = new_zisk_zkvm(resource)?;
-        run_benchmark_ere("zisk", zisk_zkvm, action, &corpuses);
+        let zisk_zkvm = new_zisk_zkvm(resource.clone())?;
```
jsign, line 175: This was a miss from a [previous PR](https://github.com/eth-act/zkevm-benchmark-workload/pull/81/files#diff-e29b59ae50fec16f88b645508dbccf764f11d9e9ee663ebca69fef48fbcb4673R175).

crates/ere-hosts/src/main.rs
```diff
@@ -165,39 +165,39 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     {
         run_cargo_patch_command("sp1")?;
         let sp1_zkvm = new_sp1_zkvm(resource.clone())?;
-        run_benchmark_ere("sp1", sp1_zkvm, action, &corpuses);
+        run_benchmark_ere("sp1", sp1_zkvm, action, &corpuses)?;
```
jsign, line 168: The reason for why `run_benchmark_ere` can fail:
- Writing the `hardware.json` failed.
- Writing benchmark json outputs failed.

Both cases shouldn't fall into the "panic handling" logic. (i.e., in the previous way we did that, if writing the json outputs failed that [was considered a crash](https://github.com/eth-act/zkevm-benchmark-workload/blob/master/crates/benchmark-runner/src/lib.rs#L106-L107) which isn't strictly correct, so improving this a bit more now)

crates/metrics/src/lib.rs
```diff
@@ -2,40 +2,131 @@
...
+        fs::write(path, json)?;
+        Ok(())
+    }
+}
+
+/// Detects available GPUs on the system.
+fn detect_gpus() -> Vec<GpuInfo> {
+    let mut gpus = Vec::new();
+
+    if let Ok(output) = std::process::Command::new("nvidia-smi")
```
jsign, line 72: If `nvidia-smi` isn't found, "no gpus" would be detected. I think for now should be fine, didn't want to get into more hairy error parsing.

Also note we only try detecting NVIDIA gpus here (compared to AMD or other). I think that's a reasonable assumption to for now.

crates/metrics/src/lib.rs
```diff
@@ -122,124 +199,128 @@ impl WorkloadMetrics {
...
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use std::iter::FromIterator;
     use tempfile::NamedTempFile;
 
     // This is just a fixed sample we are using to test serde_roundtrip
```
jsign, line 215: From here and below, mainly test fixing for new structs.

crates/metrics/src/lib.rs
```diff
@@ -2,40 +2,131 @@
...
-        name: String,
+/// Represents a single benchmark run.
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
+pub struct BenchmarkRun {
+    /// Name of the benchmark.
+    pub name: String,
+    /// Block used gas
+    pub block_used_gas: u64,
+    /// Execution metrics for the benchmark run.
+    #[serde(skip_serializing_if = "Option::is_none")]
```
kevaundray, line 16: Something to note here is that: `skip_serializing_if` doesn't play well with some encoders like bincode and rkyv when you want to derserialize the structure and the field was None.

This is okay for json, so I think its not a problem

PR #84, opened 2025-06-23
update ere and sp1 precompiles

jsign:
This PR:
- Updates `ere` to the latest release, which upgraded the `sp1-sdk` to v5.
- Updates the precompiles accordingly.


precompile-patches/sp1.toml
```diff
@@ -1,6 +1,6 @@
 [patch.crates-io]
 sha2 = { git = "https://github.com/sp1-patches/RustCrypto-hashes", tag = "patch-sha2-0.10.8-sp1-4.0.0", package = "sha2" }
```
jsign, line 2: @kevaundray, there're other precompiles available that aren't included here (e.g. `ecdsa`). 

Are the missing ones not here intentionally? Just checking that maybe we aren't missing any other one that can improve performance.

kevaundray (approved):

PR #90, opened 2025-06-27
ci: optimize checks

jsign:
This PR makes some optimizations in the CI pipeline to avoid duplicated tasks and optimize the cache usage, considering we rely heavily on feature flags.

Usual previous theoretically warm runs [take ~3.5m](https://github.com/eth-act/zkevm-benchmark-workload/actions/runs/15932909249), and [now run in 42s](https://github.com/eth-act/zkevm-benchmark-workload/actions/runs/15937628922?pr=90).

### DIFF
--- a/.github/workflows/rust_checks.yml
+++ b/.github/workflows/rust_checks.yml
@@ -20,19 +20,11 @@ jobs:
         uses: dtolnay/rust-toolchain@nightly
         with:
           components: clippy, rustfmt
-
-      - name: Cache Rust dependencies
-        uses: actions/cache@v4
+      
+      - name: Cache dependencies
+        uses: Swatinem/rust-cache@v2
         with:
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-cargo-
+          key: base 
 
       - name: Check formatting
         run: cargo fmt --check --all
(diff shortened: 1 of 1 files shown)


.github/workflows/rust_checks.yml
```diff
@@ -8,80 +8,63 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Download and Extract Fixtures
+        run: |
+          chmod +x ./scripts/download-and-extract-fixtures.sh
+          ./scripts/download-and-extract-fixtures.sh
```
jsign, line 46: We were downloading & extracting EEST fixtures many times with no good reason. Despite this was very fast, it was wasteful plus sometimes triggered a rate limiting since we also were doing it once per zkVM in the matrix below (and all at the ~same instant).

.github/workflows/rust_checks.yml
```diff
@@ -8,80 +8,63 @@ on:
...
-          path: |
-            ~/.cargo/bin/
-            ~/.cargo/registry/index/
-            ~/.cargo/registry/cache/
-            ~/.cargo/git/db/
-            target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-cargo-
+          key: ${{ matrix.zkvm }}
```
jsign, line 67: Include the zkVM as part of the cache key, so different zkVMs don't clash using the same cache since cargo would invalidate the whole cache since detected compilation flags changes which was very unfortunate.

Now each zkVM has its own cache thus subsequent runs can leverage it correctly without another zkVM destroying it.

kevaundray (approved):

PR #91, opened 2025-06-28
benchmark-runner: add flag to handle reruns & add a flag for output folder

jsign:
Yesterday, the prover we're using for running the benchmark suite crashed because of OOM. If we re-run the suite, it will start from scratch. 

This PR changes the behavior of `ere-hosts` in the following way:
- By default, if it detects that in the output folder the to-be-run case exists, it skips it and prints a message in stdout. (e.g. `Skipping tests/zkevm/test_worst_stateful_opcodes.py::test_worst_storage_access_cold[fork_Prague-blockchain_test-absent_slots_False-SSTORE same value]-2 (already exists)`)
- The user can provide a `--force-rerun` flag to override this default behavior.

I'm not sure which default behavior was the best. I suspect that if the user decides to keep an existing `zkevm-metrics` folder and rerun it, it is probably for a retry, since the most natural thing for a fresh run would be to delete the output folder (at least for me).

I also added a flag for controlling the output folder.

BTW, I suspect we'll need some opposite of `--filter`. That is, whenever we identify a benchmark case that crashes the prover, we'll probably want to use `--exclude <problematic_one>` or a similar option. But let's first reproduce the crash. (I realize now that existing `--filter` flag might not have the best name)

### DIFF
--- a/crates/benchmark-runner/src/lib.rs
+++ b/crates/benchmark-runner/src/lib.rs
@@ -4,6 +4,11 @@ use witness_generator::BlocksAndWitnesses;
 use zkevm_metrics::{BenchmarkRun, CrashInfo, ExecutionMetrics, HardwareInfo, ProvingMetrics};
 use zkvm_interface::{zkVM, Input};
 
+pub struct RunConfig {
+    pub action: Action,
+    pub force_rerun: bool,
+}
+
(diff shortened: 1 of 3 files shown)


kevaundray (approved):
LGTM -- doc comments for run config fields would be good

PR #97, opened 2025-07-01
witness-generator: accept custom EEST folder

jsign:
This PR adds a new flag in the `tests` command to provide a custom folder for EEST fixtures, since now the only option was using the script to download the latest or provided tag.

I wanted this feature since I was testing a custom filling for Osaka, and wanted to use my own EEST folder rather than using the script.

### DIFF
--- a/crates/witness-generator/src/eest_generator.rs
+++ b/crates/witness-generator/src/eest_generator.rs
@@ -1,4 +1,4 @@
-use anyhow::{Result, anyhow, bail};
+use anyhow::{Context, Result, anyhow, bail};
 use async_trait::async_trait;
 use ef_tests::{
     Case,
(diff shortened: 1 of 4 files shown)


crates/witness-generator/src/eest_generator.rs
```diff
@@ -54,27 +62,34 @@ impl ExecSpecTestBlocksAndWitnessBuilder {
...
         let tag = self.tag;
         let include = self.include.unwrap_or_default();
         let exclude = self.exclude.unwrap_or_default();
 
-        let mut cmd = Command::new("./scripts/download-and-extract-fixtures.sh");
-        if let Some(tag) = tag {
-            cmd.arg(tag);
-        }
-        let output = cmd.output()?;
+        let (directory_path, delete_eest_folder) = if let Some(input_folder) = input_folder {
```
jsign, line 70: When using a local EEST folder, the `delete_eest_folder` controls how the Drop trait works, since we don't want to delete that one since wasn't created from the script.

jsign: Yep, I'll do.

PR #100, opened 2025-07-05
witness-generator: improve testing

jsign:
This PR includes better testing coverage for the `witness-generator` crate.

### DIFF
--- a/.github/workflows/rust_checks.yml
+++ b/.github/workflows/rust_checks.yml
@@ -9,8 +9,21 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs: 
-  check:
-    name: Minimal checks on non-zkvm crates
+  check-scripts:
+    name: Check bash scripts
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Download and Extract Fixtures
+        run: |
+          chmod +x ./scripts/download-and-extract-fixtures.sh
+          ./scripts/download-and-extract-fixtures.sh
+
+  check-non-ere-hosts:
+    name: Checks on non ere-hosts crates
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
@@ -30,21 +43,8 @@ jobs:
         run: cargo fmt --check --all
 
       - name: Run tests
-        run: cargo test -p zkevm-metrics -p witness-generator -p ere-hosts
-
-  download-eest-fixtures:
-    name: Download and Extract Fixtures
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-
-      - name: Download and Extract Fixtures
-        run: |
-          chmod +x ./scripts/download-and-extract-fixtures.sh
-          ./scripts/download-and-extract-fixtures.sh
-
+        run: cargo test -p zkevm-metrics -p witness-generator
+ 
   check-ere-host:
     name: Checks crate ere-hosts
     strategy:
@@ -68,3 +68,6 @@ jobs:
 
       - name: Cargo check (with ${{ matrix.zkvm }} enabled)
         run: cargo check --features ${{ matrix.zkvm }} -p ere-hosts
+
+      - name: Run tests
+        run: cargo test -p ere-hosts
--- a/crates/witness-generator/Cargo.toml
+++ b/crates/witness-generator/Cargo.toml
@@ -23,18 +23,12 @@ tokio-util = "0.7.15"
 
 alloy-eips = { workspace = true }
 alloy-rpc-types-eth.workspace = true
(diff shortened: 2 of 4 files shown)


crates/witness-generator/Cargo.toml
```diff
@@ -23,19 +23,19 @@ tokio-util = "0.7.15"
...
 reth-stateless.workspace = true
 reth-ethereum-primitives = { workspace = true, features = [
     "serde",
     "serde-bincode-compat",
 ] }
-reth-primitives-traits = { workspace = true, features = [
-    "serde",
-    "serde-bincode-compat",
-] }
-reth-chainspec.workspace = true
```
jsign, line 37: Unused dependencies.

crates/witness-generator/src/eest_generator.rs
```diff
@@ -220,30 +222,137 @@ fn find_all_files_with_extension(path: &Path, extension: &str) -> Vec<PathBuf> {
...
-
-            let eest_generator = ExecSpecTestBlocksAndWitnesses {
-                directory_path: directory_path.clone(),
-                include: vec![],
-                exclude: vec![],
-                delete_eest_folder,
-            };
+    fn decompress_eest_release(dest_dir: &Path) -> Result<()> {
+        let path =
+            PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("testdata/zkevm_fixtures_v0.1.0.tar.gz");
```
jsign, line 233: Initially, I had a version where the script was called, so we always downloaded the whole release. But this had two problems: 1. GH rate limiting, 2. However, it is ultimately incorrect to base a test on having an internet connection.

crates/witness-generator/src/eest_generator.rs
```diff
@@ -220,30 +222,137 @@ fn find_all_files_with_extension(path: &Path, extension: &str) -> Vec<PathBuf> {
...
+        let file = File::open(path)?;
+        let buf_reader = std::io::BufReader::new(file);
+        let tar = GzDecoder::new(buf_reader);
+        let mut archive = Archive::new(tar);
+        archive.unpack(dest_dir)?;
+        Ok(())
+    }
 
-            drop(eest_generator);
+    fn prepare_downgraded_eest_fixtures(target_path: &Path) -> Result<()> {
```
jsign, line 242: This helper is built on top of the previous function, which only leaves one single EEST fixture in the folder, making witness generation fast.

Whenever we have a release that includes the downgraded 1M gas cases, _maybe_ that will be easy enough to avoid this. I'll test as soon as that setup is doable.

crates/witness-generator/src/eest_generator.rs
```diff
@@ -220,30 +222,137 @@ fn find_all_files_with_extension(path: &Path, extension: &str) -> Vec<PathBuf> {
...
+        assert!(
+            !bw_with_exclude[0].name.contains("Prague"),
+            "The fixture should not contain 'Prague' in its name"
+        );
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    #[cfg(feature = "slow-tests")]
```
jsign, line 328: This test is quite heavy since it runs witnesses for all fixtures. But as mentioned before today's only filling is for 36M which is quite heavy.

For now I'm leaving this under a `--features slow-tests` flag, and out of the CI since it's too heavy. As soon as they're downgraded for 1M baked into the release, that might be OK for GH machines and we can remove the feature. We can figure out then.

scripts/download-and-extract-fixtures.sh
```diff
@@ -4,29 +4,38 @@
 #
 # Downloads execution spec test fixtures for zkevm.
 # By default, it fetches the latest release tag starting with 'zkevm@'.
-# An optional argument can be provided to specify an exact tag.
```
jsign, line 7: When I did the tests, always pulling EEST fixtures using the script (that I reverted for the reasons mentioned before), I extended the script to also accept a target directory instead of a fixed one.

This feature isn't used in today's tests, but it remains valuable for the script.

.github/workflows/rust_checks.yml
```diff
@@ -9,8 +9,21 @@ env:
   CARGO_TERM_COLOR: always
```
jsign, line 9: A bit of a cleanup.

crates/witness-generator/src/eest_generator.rs
kevaundray: Yeah agreed, re having tests be as reproducible as possible

crates/witness-generator/src/eest_generator.rs
kevaundray: Yep, another solution might be to run these using GitHub corn jobs once a day if that's useful

crates/witness-generator/Cargo.toml
kevaundray: In a later PR, we can add a lint for this in every crate, so that unused dependencies will fail the linter and subsequently stop compilation

kevaundray (approved):

PR #101, opened 2025-07-07
witness-generator: add more tests to rpc implementation

jsign:
This PR adds more testing to the RPC implementation of `WitnessGenerator`.

### DIFF
--- a/crates/witness-generator/src/rpc_generator.rs
+++ b/crates/witness-generator/src/rpc_generator.rs
@@ -575,11 +575,11 @@ mod test {
             .await
             .expect("Failed to generate blocks and witnesses to path");
 
-        assert_eq!(
+        assert!(
             std::fs::read_dir(target_dir.path())
                 .expect("Failed to read directory")
-                .count(),
-            1,
+                .count()
+                > 0,
             "Expected at least one block"
         );
     }


crates/witness-generator/src/rpc_generator.rs
```diff
@@ -103,6 +103,13 @@ impl WitnessGenerator for RpcBlocksAndWitnesses {
     ///
     /// Returns either the last N blocks or a specific block with their execution witnesses.
     async fn generate(&self) -> Result<Vec<BlockAndWitness>> {
+        // If live polling is enabled, we return an error here
+        if self.stop.is_some() {
```
jsign, line 107: Realized this by doing the tests. ha

Under the polling configuration, given is a long running call until stopped, the only supported call now is `generate_to_path` since results are constantly saved in the files.

I think in the future we can support returning the result in a `Vec` but I'm not sure this will be used by somebody, since if you're doing really long runs (as we do in our daemon VM), that would be a huge vector. But maybe for smaller time spans might make sense.

.github/workflows/rust_checks.yml
```diff
@@ -43,7 +43,10 @@ jobs:
         run: cargo fmt --check --all
 
       - name: Run tests
-        run: cargo test -p zkevm-metrics -p witness-generator
+        env:
+          RPC_URL: ${{ secrets.RPC_URL }}
+          RPC_HEADERS: ${{ secrets.RPC_HEADERS }}
```
jsign, line 48: Added these as repository secrets

.github/workflows/rust_checks.yml
```diff
@@ -43,7 +43,10 @@ jobs:
...
+        run: cargo test -p zkevm-metrics -p witness-generator -- --no-capture
```
jsign, line 49: The `--no-capture` is for double-checking the test aren't skipped due to envs not properly set. I did a test of not setting the env vars to double check that all was running as expected.

crates/witness-generator/src/rpc_generator.rs
```diff
@@ -417,3 +424,163 @@ impl TryFrom<RpcFlatHeaderKeyValues> for HeaderMap {
...
+            return;
+        }
+
+        let stop_token = CancellationToken::new();
+
+        // Spawn a task to cancel the token after ~12s which should be enough time to fetch at least one block.
+        tokio::spawn({
+            let stop_token = stop_token.clone();
+            async move {
+                tokio::time::sleep(std::time::Duration::from_secs(12)).await;
```
jsign, line 561: I was tempted to mark this "slow", but for now leaning on the safe side and always running it. Another strategy is avoid the 12s wait and inspect for a file appearing and stop as soon as possible, but still not sure that changes thing much (maybe safe 6s on avg). Ended up on the simpler side.

jsign:
> LGTM -- In the future, we likely want to mock out the RPC to avoid flaky tests and a dependency on live RPC servers (if the env variable is set)

Agree, or maybe have both modes (mocked & real) and run the real live RPC nightly or something -- mostly to double check that the API response in a real node hasn't changed and things are broken.

PR #107, opened 2025-07-11
Separate host and guest workspace so guest patching does not affect host

han0110:
- Separate host and guest workspace so guest patching does not affect host
  - The goods:
    - When we are developing host crates, we don't need to reset the patching frequently.
    - The rust-analyzer is not that busy
  - The bads: the `target` dir now becomes 2, but the guest one rarely changes so I think it should be fine.
- Refactos the checks that at least one feature has to be turned on to compile time.


jsign (commented):
@han0110, I think makes sense! cc @kevaundray  if there's any pusback on this.

Reg the CI, I think is failing since the PR comes from a fork so secrets aren't available for security reasons, so it should be OK.

PR #110, opened 2025-07-14
chore: update reth

kevaundray:
Changes in fork: Merged main into the branch we are using

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -231,9 +231,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-evm"
-version = "0.12.3"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


jsign:
> @jsign should the bash script be looking for the benchmark prefiix instead of the zkevm prefix now?

No, we can't since the `benchmark@...` releases have an unsupported format. I detected this a while ago when I wanted to do the prefix change, and is [tracked here](https://github.com/ethereum/execution-spec-tests/issues/1778). As soon as [this PR](https://github.com/ethereum/execution-spec-tests/pull/1855) gets merged and a new tag is done, I'll update the prefix in the script.

jsign:
BTW, the CI failing is GH rate limiting. I think for CI will be better to download the releases to the repo and use them locally to avoid this. We can do that whenever the next release of `benchmark@...` is done -- meantime is just retrying failed jobs.

jsign (approved):

PR #114, opened 2025-07-15
Update ere using Docker for some guest program compilations

jsign:
See https://github.com/eth-act/ere/pull/54

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,8 +2,8 @@
 CARGO_WORKSPACE_DIR = { value = "", relative = true }
 
 [alias]
-pico = ["run", "-p", "xtask", "--", "pico", "--"]
-risc0 = ["run", "-p", "xtask", "--", "risc0", "--"]
-sp1 = ["run", "-p", "xtask", "--", "sp1", "--"]
-zisk = ["run", "-p", "xtask", "--", "zisk", "--"]
-zkm = ["run", "-p", "xtask", "--", "zkm", "--"]
+pico = ["run", "-p", "xtask", "--", "pico"]
+risc0 = ["run", "-p", "xtask", "--", "risc0"]
+sp1 = ["run", "-p", "xtask", "--", "sp1"]
+zisk = ["run", "-p", "xtask", "--", "zisk"]
+zkm = ["run", "-p", "xtask", "--", "zkm"]
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,7 +1,7 @@
 name: Integration tests
 on:
   push:
-    branches: [ main, master ]
+    branches: [ main, master, jsign-ere-docker ]
     schedule:
       - cron: '0 0 * * *'
   workflow_dispatch:
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ zkevm-metrics
 
 # private envrc file
 .envrc
+
+ere-guests/**/Cargo.lock
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -105,12 +105,12 @@ witness-generator = { path = "crates/witness-generator" }
 zkevm-metrics = { path = "crates/metrics" }
 benchmark-runner = { path = "crates/benchmark-runner" }
 
-zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", tag = "v0.0.5", package = "zkvm-interface" }
-ere-sp1 = { git = "https://github.com/eth-applied-research-group/ere", tag = "v0.0.5", package = "ere-sp1" }
-ere-risczero = { git = "https://github.com/eth-applied-research-group/ere", tag = "v0.0.5", package = "ere-risczero" }
-ere-pico = { git = "https://github.com/eth-applied-research-group/ere", tag = "v0.0.5", package = "ere-pico" }
-ere-openvm = { git = "https://github.com/eth-applied-research-group/ere", tag = "v0.0.5", package = "ere-openvm" }
-ere-zisk = { git = "https://github.com/eth-applied-research-group/ere", tag = "v0.0.5", package = "ere-zisk" }
+zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", rev = "82e9f9aae06f045eda21d79940aaf1f1981265b1", package = "zkvm-interface" }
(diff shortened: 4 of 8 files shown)


.cargo/config.toml
```diff
@@ -2,8 +2,8 @@
...
-pico = ["run", "-p", "xtask", "--", "pico", "--"]
-risc0 = ["run", "-p", "xtask", "--", "risc0", "--"]
-sp1 = ["run", "-p", "xtask", "--", "sp1", "--"]
-zisk = ["run", "-p", "xtask", "--", "zisk", "--"]
-zkm = ["run", "-p", "xtask", "--", "zkm", "--"]
+pico = ["run", "-p", "xtask", "--", "pico"]
+risc0 = ["run", "-p", "xtask", "--", "risc0"]
+sp1 = ["run", "-p", "xtask", "--", "sp1"]
+zisk = ["run", "-p", "xtask", "--", "zisk"]
+zkm = ["run", "-p", "xtask", "--", "zkm"]
```
jsign, line 9: I removed this since now `xtask` accepts an optional flag `--manifest-folder`.
For SP1, this is set explicitly since we want to target a specific guest program crate instead of the workspace. If not provided, it defaults to the `ere-guest` workspace (same as before).

The `--` now is mainly a responsibility of whoever wants to pass arguments to cargo (if any, I don't think we rely on this today).

I think at some point might be good to do the patching via libs instead of `cargo -p xtasks`? As in, `ere-hosts` using `xtasks` as a library instead of via `cargo`.

In any case, no need to really do that if mid-term the way patching works works and we don't need to do them in crates (it might take a while though). 

No strong opinions -- I think we can leave it this way.

.github/workflows/integration-tests.yml
```diff
@@ -1,7 +1,7 @@
 name: Integration tests
 on:
   push:
-    branches: [ main, master ]
+    branches: [ main, master, jsign-ere-docker ]
```
jsign, line 4: Temporal change to double check integration tests, I'll remove before merging.

.gitignore
```diff
@@ -10,4 +10,6 @@ zkevm-metrics
 .vscode
 
 # private envrc file
-.envrc
\ No newline at end of file
+.envrc
+
+ere-guests/**/Cargo.lock
```
jsign, line 15: I think is a good idea to avoid guest programs `Cargo.lock` being interesting for this repo git?

Cargo.toml
```diff
@@ -105,12 +105,12 @@ witness-generator = { path = "crates/witness-generator" }
...
-ere-openvm = { git = "https://github.com/eth-applied-research-group/ere", tag = "v0.0.5", package = "ere-openvm" }
-ere-zisk = { git = "https://github.com/eth-applied-research-group/ere", tag = "v0.0.5", package = "ere-zisk" }
+zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", rev = "82e9f9aae06f045eda21d79940aaf1f1981265b1", package = "zkvm-interface" }
+ere-sp1 = { git = "https://github.com/eth-applied-research-group/ere", rev = "82e9f9aae06f045eda21d79940aaf1f1981265b1", package = "ere-sp1" }
+ere-risczero = { git = "https://github.com/eth-applied-research-group/ere", rev = "82e9f9aae06f045eda21d79940aaf1f1981265b1", package = "ere-risczero" }
+ere-pico = { git = "https://github.com/eth-applied-research-group/ere", rev = "82e9f9aae06f045eda21d79940aaf1f1981265b1", package = "ere-pico" }
+ere-openvm = { git = "https://github.com/eth-applied-research-group/ere", rev = "82e9f9aae06f045eda21d79940aaf1f1981265b1", package = "ere-openvm" }
+ere-zisk = { git = "https://github.com/eth-applied-research-group/ere", rev = "82e9f9aae06f045eda21d79940aaf1f1981265b1", package = "ere-zisk" }
```
jsign, line 113: Temp patching until https://github.com/eth-act/ere/pull/54 is merged and we create a new tag. Will change before merging.

crates/ere-hosts/src/main.rs
```diff
@@ -194,7 +194,7 @@ fn get_zkvm_instances(
     {
         #[cfg(feature = "sp1")]
         {
-            run_cargo_patch_command("sp1")?;
+            run_cargo_patch_command("sp1", Some(guest_program_folder))?;
```
jsign, line 197: `run_cargo_patch_command` now accepts an optional guest program dir where you want to do the crate patching.

ere-guests/stateless-validator/sp1/Cargo.toml
```diff
@@ -1,32 +1,35 @@
...
 ] }
-reth-primitives-traits = { workspace = true, features = [
+reth-primitives-traits = { git = "https://github.com/kevaundray/reth", rev = "03364a836774c72f4e354de924330fee6a41be68", features = [
     "serde",
     "serde-bincode-compat",
 ] }
-reth-evm-ethereum = { workspace = true }
-reth-chainspec = { workspace = true }
+reth-evm-ethereum = { git = "https://github.com/kevaundray/reth", rev = "03364a836774c72f4e354de924330fee6a41be68" }
+reth-chainspec = { git = "https://github.com/kevaundray/reth", rev = "03364a836774c72f4e354de924330fee6a41be68" }
```
jsign, line 21: This was the "ugly" part of this story, since should be in sync with:
- `ere_hosts` since it is the one generating the proofs. So ideally guest programs SDK match the prover version.
- `ere` since it is the one installing the compilation chain for the guest program

@kevaundray, [mentioned](https://github.com/eth-act/ere/pull/54#issuecomment-3074353585) some potential idea to potentially generalize the way guest programs are created to reduce the burden. e.g., all `stateless-validator` guest program use the same `reth-*` dependencies and main guest program implementation.

jsign:
@kevaundray, maybe you want to take a look and see how this feels.
If looks good, I can proceed with the OP list of tasks.

jsign:
@kevaundray, updated with latest `ere` tag -- should be ready!

.github/workflows/integration-tests.yml
```diff
@@ -56,14 +56,6 @@ jobs:
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@nightly
 
-      - name: Install SP1
```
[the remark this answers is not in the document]
kevaundray: Nice!

PR #117, opened 2025-07-17
ere-guests: add rlp encoding length

jsign:
This PR adds a new guest program type `rlp-encoding-length`. 

The goal is to measure the potential impact of [EIP-7934](https://eips.ethereum.org/EIPS/eip-7934).

The `benchmark-runner` has a pending refactor pending to reorganize a bit some stuff to avoid repetition. I preferred not to do that in this PR to avoid bloating it and keep it focused. With this PR is already at a point where I think that should be the next PR I'll will be working on after, since more guest program types are coming.

### DIFF
--- a/ere-guests/rlp-encoding-length/sp1/Cargo.toml
+++ b/ere-guests/rlp-encoding-length/sp1/Cargo.toml
@@ -6,17 +6,18 @@ rust-version = "1.85"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-reth-ethereum-primitives = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf", features = [
-    "serde",
-    "serde-bincode-compat",
-] }
-
 sp1-zkvm = "5.0.5"
-alloy-consensus = { version = "1.0.18", default-features = false }
-reth-primitives-traits = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf" }
+
 tracing = "*"
 tracing-subscriber = "*"
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_with = "3"
+
+reth-ethereum-primitives = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf", features = [
+    "serde",
+    "serde-bincode-compat",
+] }
+alloy-consensus = { version = "1.0.18", default-features = false }
+reth-primitives-traits = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf" }
 
 [patch.crates-io]


ere-guests/rlp-encoding-length/sp1/src/main.rs
```diff
@@ -0,0 +1,20 @@
...
+#![no_main]
+
+use alloy_consensus::Block;
+
+sp1_zkvm::entrypoint!(main);
+/// Entry point.
+pub fn main() {
+    println!("cycle-tracker-report-start: read_input");
+    let block = sp1_zkvm::io::read::<reth_stateless::BincodeBlock>();
+    let iterations = sp1_zkvm::io::read::<u16>();
```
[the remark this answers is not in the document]
jsign: BTW, from what I see Reth doesn't "cache" this calculation in the way is implemented.
I think other clients or implementations might do that -- but doesn't seem to be the case for Reth (if that would be the case, this `iterations` intention would be borked)

ere-guests/rlp-encoding-length/sp1/src/main.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#![no_main]
+
+use alloy_consensus::Block;
+
+sp1_zkvm::entrypoint!(main);
+/// Entry point.
+pub fn main() {
+    println!("cycle-tracker-report-start: read_input");
+    let block = sp1_zkvm::io::read::<reth_stateless::BincodeBlock>();
```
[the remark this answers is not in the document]
jsign: @kevaundray, BTW, I could also avoid the reth-fork change and do the wrapper on our repository. I would have to pull a bunch of new reth crates but can also be more transparent.

kevaundray (approved):

ere-guests/rlp-encoding-length/sp1/src/main.rs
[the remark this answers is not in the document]
jsign: Yep, agree. I'll define that type in a new crate lib here so we don't leave hanging stuff we know we have to remove in the reth-fork.

Thanks!

kevaundray:
One step closer to renaming this repo to `zkvm-benchmark-workload` :)

Cargo.toml
```diff
@@ -114,10 +114,10 @@ ere-zisk = { git = "https://github.com/eth-applied-research-group/ere", tag = "v
...
 # branch is kw/zkevm-benchmark-workload-repo
 # NOTE: We are using a branch of a branch that has not yet been merged into master.
-ef-tests = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf" }
-reth-ethereum-primitives = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf" }
-reth-stateless = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf" }
-reth-rpc-api = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf" }
+ef-tests = { git = "https://github.com/kevaundray/reth", rev = "246689b8caf148f4185c5a19d90b9c15de12ed89" }
+reth-ethereum-primitives = { git = "https://github.com/kevaundray/reth", rev = "246689b8caf148f4185c5a19d90b9c15de12ed89" }
+reth-stateless = { git = "https://github.com/kevaundray/reth", rev = "246689b8caf148f4185c5a19d90b9c15de12ed89" }
+reth-rpc-api = { git = "https://github.com/kevaundray/reth", rev = "246689b8caf148f4185c5a19d90b9c15de12ed89" }
```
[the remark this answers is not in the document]
jsign: This won't be done, and we'll include the type in this repo.

Cargo.toml
```diff
@@ -3,9 +3,11 @@ members = [
     "crates/metrics",
     "crates/witness-generator",
     "crates/witness-generator-cli",
-    "xtask",
-    # Ere host crate
     "crates/ere-hosts",
+
+    "ere-guests/libs",
```
jsign, line 8: Starting to prepare the stage for the shared lib to be used for guests and required stdin types.

Cargo.toml
```diff
@@ -118,10 +121,13 @@ ef-tests = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311d
 reth-ethereum-primitives = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf" }
 reth-stateless = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf" }
 reth-rpc-api = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf" }
+reth-primitives-traits = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf" }
```
jsign, line 124: I moved back to our current reth-fork, so now creating the `BincodeBlock` in this new `ere-guests/lib` crate.

The idea is that this crate will be part of both the main workspace of this repo, and the `ere-guests` workspace. But the latter will happen when we do the `ere`-Dockerization change.

crates/benchmark-runner/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 
+use guest_libs::BincodeBlock;
```
jsign, line 5: Much nicer.

ere-guests/libs/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
```
jsign, line 1: The new crate.

ere-guests/rlp-encoding-length/sp1/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+tracing-subscriber = "*"
+serde = { version = "1.0", default-features = false, features = ["derive"] }
+serde_with = "3"
+
+reth-ethereum-primitives = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf", features = [
+    "serde",
+    "serde-bincode-compat",
+] }
+alloy-consensus = { version = "1.0.18", default-features = false }
+reth-primitives-traits = { git = "https://github.com/kevaundray/reth", rev = "75092d6aa098311db96c06b97a252a6f2b42dbaf" }
```
jsign, line 21: This is temporary until we support thew new `ere`-Dockerization, since now guests must be self contained.

After we do the change, we'll replace all these dependencies with `guest-libs.workspace = true`.

ere-guests/rlp-encoding-length/sp1/src/main.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub Block,
+);
+
+impl Deref for BincodeBlock {
+    type Target = Block;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
```
jsign, line 43: and this also will go away since we can use `guest-libs.WrappedBlock` type directly.

jsign (commented):
@kevaundray, done some extra changes reg not requiring a new reth-fork.

I resolved old comments, and created new ones only for new changes for a quick last review.

PR #119, opened 2025-07-19
guest-programs: use new workspace mounting from ere

jsign:
This PR updates to a new `ere` version which allows to distinguish between:
- Workspace directory: which directory gets mounted at compilation time.
- Guest program directory: a _relative_ path within the workspace to be compiled.

The main benefits are:
- We can switch again guest program types for zkVMs to use workspace dependencies, which reduces the burden of matching dependency versions.
- Guest programs can now use the `guest-libs` as a dependency, which will help reduce repetition in all guest program types.

As a first example of the latter, the `BincodeBlock` type that we had to duplicate before is now shared between `ere-hosts` (which is the "sender" of the stdin) and the guest program (which is the receiver).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,12 +108,12 @@ zkevm-metrics = { path = "crates/metrics" }
 benchmark-runner = { path = "crates/benchmark-runner" }
 guest-libs = { path = "ere-guests/libs" }
 
-zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", rev = "9444a2ba73ba933b573b80d9c66c4fbc3bcbe024", package = "zkvm-interface" }
-ere-sp1 = { git = "https://github.com/eth-applied-research-group/ere", rev = "9444a2ba73ba933b573b80d9c66c4fbc3bcbe024", package = "ere-sp1" }
-ere-risczero = { git = "https://github.com/eth-applied-research-group/ere", rev = "9444a2ba73ba933b573b80d9c66c4fbc3bcbe024", package = "ere-risczero" }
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -108,12 +108,12 @@ zkevm-metrics = { path = "crates/metrics" }
...
-ere-openvm = { git = "https://github.com/eth-applied-research-group/ere", tag = "v0.0.7", package = "ere-openvm" }
-ere-zisk = { git = "https://github.com/eth-applied-research-group/ere", tag = "v0.0.7", package = "ere-zisk" }
+zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", rev = "1bfacf473846cb6115382b2a2d23d6cc9f02546e", package = "zkvm-interface" }
+ere-sp1 = { git = "https://github.com/eth-applied-research-group/ere", rev = "1bfacf473846cb6115382b2a2d23d6cc9f02546e", package = "ere-sp1" }
+ere-risczero = { git = "https://github.com/eth-applied-research-group/ere", rev = "1bfacf473846cb6115382b2a2d23d6cc9f02546e", package = "ere-risczero" }
+ere-pico = { git = "https://github.com/eth-applied-research-group/ere", rev = "1bfacf473846cb6115382b2a2d23d6cc9f02546e", package = "ere-pico" }
+ere-openvm = { git = "https://github.com/eth-applied-research-group/ere", rev = "1bfacf473846cb6115382b2a2d23d6cc9f02546e", package = "ere-openvm" }
+ere-zisk = { git = "https://github.com/eth-applied-research-group/ere", rev = "1bfacf473846cb6115382b2a2d23d6cc9f02546e", package = "ere-zisk" }
```
jsign, line 116: Using https://github.com/eth-act/ere/pull/57. Waiting for review, merge & tag to edit this again.

PR #120, opened 2025-07-22
guest-programs: support SSZ encoding length

jsign:
This PR adds support for benchmarking block SSZ encoding.

In the context of [EIP-7934](https://eips.ethereum.org/EIPS/eip-7934), in [previous PR](https://github.com/eth-act/zkevm-benchmark-workload/pull/117) we introduced a guest program that calculates the RLP length of blocks. This PR adds SSZ format support to understand how this compares to a theoretical SSZ encoding reality (although isn't part of EIP-7934).

Some notes on the implementation:
- There isn't native SSZ support for `alloy` types. Compared to having another fork of `alloy` and I leaned towards having our own minimal types for proper `ethereum_ssz` automatic derivation of `ssz::{Encode | Decode}`. This is quite verbose, so I created a separate module in `guest-libs` for this -- whenever `alloy` has native support prob via a new feature, we can nuke that module and should be fine.
- Initially, I created a separate guest program for the SSZ-encoding-length but later realized this creates a lot of repetition in `ere-hosts` for no good reason -- so I generalized the existing `rlp-encoding-length` guest program to `block-encoding-length` which accepts an extra argument with the desired format.

I ran many mainnet blocks, both with RLP and SSZ, and everything looks good. I need to run it in a larger set, and we can conduct some analysis to compare the results.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1658,7 +1658,7 @@ dependencies = [
 [[package]]
 name = "build-utils"
 version = "0.1.0"
-source = "git+https://github.com/eth-applied-research-group/ere?rev=v0.0.8#f05aa500327fd130582b2732ce0e2c7a64e74214"
+source = "git+https://github.com/eth-applied-research-group/ere?rev=v0.0.9#498c484d1c8ad971b2afe8735e0a42f346b2cf92"
 dependencies = [
  "cargo_metadata 0.20.0",
  "thiserror 2.0.12",
@@ -3110,7 +3110,7 @@ dependencies = [
 [[package]]
 name = "ere-openvm"
 version = "0.1.0"
-source = "git+https://github.com/eth-applied-research-group/ere?rev=v0.0.8#f05aa500327fd130582b2732ce0e2c7a64e74214"
+source = "git+https://github.com/eth-applied-research-group/ere?rev=v0.0.9#498c484d1c8ad971b2afe8735e0a42f346b2cf92"
 dependencies = [
  "build-utils",
  "openvm-build",
@@ -3125,7 +3125,7 @@ dependencies = [
 [[package]]
 name = "ere-pico"
 version = "0.1.0"
-source = "git+https://github.com/eth-applied-research-group/ere?rev=v0.0.8#f05aa500327fd130582b2732ce0e2c7a64e74214"
+source = "git+https://github.com/eth-applied-research-group/ere?rev=v0.0.9#498c484d1c8ad971b2afe8735e0a42f346b2cf92"
 dependencies = [
  "bincode",
  "build-utils",
@@ -3137,7 +3137,7 @@ dependencies = [
 [[package]]
 name = "ere-risczero"
 version = "0.1.0"
-source = "git+https://github.com/eth-applied-research-group/ere?rev=v0.0.8#f05aa500327fd130582b2732ce0e2c7a64e74214"
+source = "git+https://github.com/eth-applied-research-group/ere?rev=v0.0.9#498c484d1c8ad971b2afe8735e0a42f346b2cf92"
 dependencies = [
  "anyhow",
  "borsh",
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -125,8 +125,11 @@ This repository contains an `xtask` that will automate this process by calling `
     # Run empty program benchmarks (for measuring zkVM overhead)
     cargo run --release --features sp1 -- empty-program
     
-    # Run RLP encoding length benchmarks
-    cargo run --release --features sp1 -- rlp-encoding-length --loop-count 100
+    # Run block encoding length benchmarks (with RLP encoding format)
+    cargo run --release --features sp1 -- block-encoding-length --loop-count 100 --format rlp
+    
+    # Run block encoding length benchmarks (with SSZ encoding format)
+    cargo run --release --features sp1 -- block-encoding-length --loop-count 100 --format ssz
```
jsign, line 132: TL;DR on how the guest program was generalized to the encoding format.

crates/benchmark-runner/src/guest_programs.rs
```diff
@@ -64,29 +64,43 @@ pub fn stateless_validator_generate_inputs(
...
             let mut stdin = Input::new();
-            let metadata = RLPLengthMetadata {
+            let metadata = BlockEncodingLengthMetadata {
+                format: format!("{format:?}"),
                 block_hash: bw.block_and_witness.block.hash_slow().to_string(),
                 loop_count,
             };
             stdin.write(BincodeBlock(bw.block_and_witness.block));
             stdin.write(loop_count);
+            stdin.write(format as u8);
```
jsign, line 103: A new input to the guest program with the format.

ere-guests/block-encoding-length/sp1/src/main.rs
```diff
@@ -0,0 +1,43 @@
...
+        }
+        SSZ_FORMAT => {
+            println!("cycle-tracker-report-start: block_format_conversion");
+            let block: block_ssz::Block = block.0.into();
+            println!("cycle-tracker-report-end: block_format_conversion");
+            println!("cycle-tracker-report-start: block_encoding_length_calculation");
+            for _ in 0..iterations {
+                block.ssz_bytes_len();
+            }
+            println!("cycle-tracker-report-end: block_encoding_length_calculation");
```
jsign, line 39: The new feat

crates/benchmark-runner/src/guest_programs.rs
```diff
@@ -64,29 +64,43 @@ pub fn stateless_validator_generate_inputs(
     Ok(guest_inputs)
 }
 
-/// Metadata for the block RLP length calculation guest program.
+/// Metadata for the block block length calculation guest program.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct RLPLengthMetadata {
+pub struct BlockEncodingLengthMetadata {
+    format: String,
```
jsign, line 70: Since the guest program was generalized, now the encoding format is part of the metadata to know by looking at the results which format was used for the run and potential comparisons.

ere-guests/libs/src/block_ssz.rs
```diff
@@ -0,0 +1,568 @@
...
+        }
+    }
+}
+
+/// SSZ-serializable transaction envelope supporting a subset of Ethereum transaction types.
+#[derive(Debug, PartialEq, Eq, ssz_derive::Encode, ssz_derive::Decode)]
+#[ssz(enum_behaviour = "union")]
+pub enum EthereumTxEnvelope {
+    /// Legacy transaction type.
+    Legacy(SignedTx<TxLegacy>),
```
jsign, line 151: First I wanted to only support legacy transaction since is the only type used in EEST -- so I didn't had to bloat this module with everything.

But later I realized that for realistic  benchmark comparisons we would like to run with mainnet blocks -- doing an early run realized I needed all, so had to go "all in". NBD, but just a bit more bloat.

ere-guests/libs/src/block_ssz.rs
jsign, line 1: This file is quite long and really not interesting:
- Structs that are mostly copy-pastes of `alloy` ones but without generics (e.g., we use Ethereum `Header` since we aren't interested of "generalized chain headers", and similar).
- I still use generics in `SignedTx<...>` but that's fine for `ssz_derive` since those types always are `ssz::{Encode | Decode}`.
- Do all `From<alloy::<corresponding type>` to transparently transform the block received in the guest stdin into this SSZ-friendly types.

As I added in this comment -- this module is somewhat unfortunate, but I think as soon as there's SSZ support in Reth/alloy, then we can come here and nuke the entire module and I think should be fine.

kevaundray (approved):
LGTM -- I think soon it would be good to start thinking about how to make the adding of guest programs scale -- There are two things which might introduce more architectural changes that we may want to wait a little bit for: 

-  integrating all of the docker stuff
-  having a non-rust program being used

PR #125, opened 2025-07-29
ci: use custom gh runners

jsign:


.github/workflows/integration-tests.yml
```diff
@@ -42,7 +52,7 @@ jobs:
   
   stateless-validator:
```
jsign, line 53: BTW, tried including `prove` in `action: [execute]` but it just takes hours. I had the sensation this wasn't that heavy, but actually tried in my own machine and CPU proving is really that slow.

.github/workflows/integration-tests.yml
```diff
@@ -21,12 +21,22 @@ jobs:
         with:
           components: clippy, rustfmt
       
+      - name: Cache dependencies
```
jsign, line 24: Doesn't look like the runners really persist this cache which is somewhat unfortunate, since the compilation with `--release` takes a while. But still looks tractable.

jsign:
~15m feels a bit borderline, but honestly I would sleep better if this runs on every PR. :)
We can eventually ask for a beefier machine, but I've been asking too many things lately to the devops team :smile: 

cc @kevaundray @han0110

PR #127, opened 2025-08-01
stateless-validator: Use optimized MPT

jsign:
This PR makes the `stateless-validator` use as the MPT implementation [coded by Risc0](https://github.com/risc0/risc0-ethereum/tree/main/crates/trie) instead of the [default in Reth](https://github.com/kevaundray/reth/blob/main/crates/stateless/src/trie.rs). Using external MPT implementations in Reth is possible thanks to the inclusion of [this trait](https://github.com/paradigmxyz/reth/pull/17098) in the STF function.

I ran a benchmark in 100 consecutive mainnet blocks, and here’s a TL;DR — more details are provided later:

- Avg. speedup of *total block* cycle counts: 1.62x
- Avg. speedup of raw block proving time (L40s): 1.44x

Summary regarding cycles ([more details here](https://gist.githubusercontent.com/jsign/0bfb5fb0196ccbaef74691e6fbf8c3c0/raw/56740e2de4fd35b44dbff053bd1269c4d78df75a/cmp)):

```
Regions ranked by average speedup:
  1. Post State Compute: 4.83x (+383.1%) - IMPROVEMENT
  2. Verify Witness: 3.10x (+209.8%) - IMPROVEMENT
  3. Validation: 1.64x (+63.6%) - IMPROVEMENT
  4. Recover Signers: 1.00x (+0.0%) - NO CHANGE
  5. Read Input: 1.00x (+0.0%) - NO CHANGE
  6. Decode Headers: 1.00x (+0.0%) - NO CHANGE
  7. Block Execution: 0.76x (-23.7%) - REGRESSION

Total files analyzed: 100

🎯 OVERALL PERFORMANCE (Total Num Cycles): 1.62x (+61.7%) - IMPROVEMENT
```

The *Verify witness* and *Post state compute* have very nice speedups, but should be interpreted carefully since there’s a slowdown in *Block execution*. The reason for this is that the Risc0 MPT implementation lazily loads the storage tries of accounts on first access, unlike the Reth implementation, which does all the deserialization before the block executes. Both approaches should be fine under the assumption that the witness is canonical (if that isn’t the case, Risc0 approach would be better to avoid wasted effort). In any case, the important part is the *Total Num Cycles* speedup.

Summary regarding wall-clock proving time ([more details here](https://gist.githubusercontent.com/jsign/abe7eed122b0c75cc67ef103154d9e37/raw/605b470402b2fcb82d10d6ad3505ee4d2c5b9e3b/cmp_proving)):

```
OVERALL SUMMARY:

PROVING TIME PERFORMANCE:
  Average speedup: 1.44x (+44.4%)
  Min speedup: 1.33x
  Max speedup: 1.54x
  Total baseline time: 41,968 seconds
  Total optimized time: 29,139 seconds
  Total time saved: 12,828 seconds
  Top 3 best speedups:
    1. sp1-v5.1.0/rpc_block_22974657: 1.54x (saved 108 s)
    2. sp1-v5.1.0/rpc_block_22974575: 1.52x (saved 173 s)
    3. sp1-v5.1.0/rpc_block_22974632: 1.52x (saved 166 s)
  Top 3 worst speedups:
    1. sp1-v5.1.0/rpc_block_22974663: 1.35x (saved 141 s)
    2. sp1-v5.1.0/rpc_block_22974580: 1.34x (saved 138 s)
    3. sp1-v5.1.0/rpc_block_22974640: 1.33x (saved 118 s)

================================================================================
KEY FINDINGS
================================================================================

Total files analyzed: 100

🎯 OVERALL PROVING PERFORMANCE: 1.44x (+44.4%) - IMPROVEMENT
   Total time saved: 12,828 seconds
   Efficiency gain: 30.7% reduction in proving time
```

As mentioned, the avg speedup is 1.44x. Note that the top 3 best and worst speedups are uncorrelated to gas usage (if you check in Etherscan). This makes sense, since the speedup is buried in the details on how much gas was used for storage access, plus the potential ratio of bytecode usage vs trie usage.

(Lateral comment getting into the weeds: “verifying the witness” has work related to the MPT (which is what this optimized impl helps with), but also has the keccak hashing of all the bytecodes, which is unrelated to the MPT implementation. This part can’t be optimized unless bytecodes are chunkified — I think at some point it could be interesting to unpack a bit how many cycles of the witness verification are related to bytecode hashing. This might only be relevant for avg. block analysis, worst/killer cases are already known)


Cargo.toml
```diff
@@ -131,10 +133,15 @@ alloy-consensus = { version = "1.0.18", default-features = false }
...
 alloy-chains = { version = "0.2.5", default-features = false }
+alloy-rlp = { version = "0.3.10", default-features = false }
+alloy-trie = { version = "0.9.0", default-features = false }
 
 ethereum_ssz_derive = "0.9"
 ethereum_ssz = "0.9"
 
+revm-bytecode = { version = "6.0.1", default-features = false }
+
+risc0-ethereum-trie = { git = "https://github.com/risc0/risc0-ethereum.git", rev = "4a5efe13b3655258b89e7082c57743e445133dc8" }
```
jsign, line 144: Sticking to the [same reference](https://github.com/risc0/zeth/blob/main/Cargo.toml#L20)

kevaundray: Can we create a tracking issue to eventually remove this (the MPT impl) when/if it gets added to upstream?

kevaundray (approved):

PR #129, opened 2025-08-01
ci: bump rust version for witness-generator-cli image generation

jsign:
Fixes `witness-generation-cli` CI [failure](https://github.com/eth-act/zkevm-benchmark-workload/actions/runs/16680811518/job/47218771788) that only runs in `master`.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -21,11 +21,6 @@ jobs:
         with:
           components: clippy, rustfmt
       
-      - name: Cache dependencies
-        uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -21,11 +21,6 @@ jobs:
         with:
           components: clippy, rustfmt
       
-      - name: Cache dependencies
```
jsign, line 24: Unrelated to main PR intention, but unfortunately, I've to remove this. In the devops GH runners, pulling the cache works at ~1MiB/s, and rust caches are quite big so it's quite slow.

image

PR #138, opened 2025-08-11
Use ChainConfig for stateless-validator input

jsign:
This PR uses `ChainConfig` as the field passed to the stateless validator for block validation. 

The PR currently uses a Reth-fork commit from [this PR](https://github.com/kevaundray/reth/pull/20). I recommend looking at that PR first to know the changes on the other side.

Examples of running `witness-generator-cli` with `ChainConfig`:

- [EEST](https://gist.github.com/jsign/0afd660f09294496dd3b1166c7fb1aa0#file-eest_with_chainconfig-json-L99-L134)
- [RPC](https://gist.github.com/jsign/c0573a0aa4fd89ab0af70abe9ed1c9f5#file-mainnet_with_chainconfig-json-L5-L40) (deleted some known fields since it was too big for GH gist)

Considering passing `ChainConfig` is a bigger structure than the previous `ForkSpec` it is worth comparing if there’s any performance hit. The following is the summary of a benchmark run of 100 consecutive mainnet blocks ([full output](https://gist.github.com/jsign/70bc542b0a14284630bd02ba97564eae)):
```
================================================================================
KEY FINDINGS
================================================================================

Regions ranked by average speedup:
  1. Verify Witness: 1.00x (+0.1%) - NO CHANGE
  2. Validation: 1.00x (+0.0%) - NO CHANGE
  3. Post State Compute: 1.00x (+0.0%) - NO CHANGE
  4. Block Execution: 1.00x (+0.0%) - NO CHANGE
  5. Recover Signers: 1.00x (+0.0%) - NO CHANGE
  6. Decode Headers: 1.00x (+0.0%) - NO CHANGE
  7. Read Input: 0.99x (-1.0%) - REGRESSION

Total files analyzed: 100

🎯 OVERALL PERFORMANCE (Total Num Cycles): 1.00x (+0.0%) - NO CHANGE
```
This means there's a ~1% regression in `read_inputs` but in the grand scheme of things `total_num_cycles` is negligible.

This PR also keeps a helper script to migrate the old format of collected witnesses to the new format. This was useful to run the same gathered 100 mainnet blocks and run them in both branches, which is going to be useful in the future because we have already collected a lot of mainnet blocks that will need to be migrated eventually.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2917,7 +2917,7 @@ dependencies = [
 [[package]]
 name = "ef-tests"
 version = "1.6.0"
-source = "git+https://github.com/kevaundray/reth?rev=f17c70c23056cadd7f42ee7c4502f72078d2a118#f17c70c23056cadd7f42ee7c4502f72078d2a118"
+source = "git+https://github.com/kevaundray/reth?rev=0880801d0d6c29d1a7daaee24718b73e5a8a52a8#0880801d0d6c29d1a7daaee24718b73e5a8a52a8"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -8823,7 +8823,7 @@ checksum = "95325155c684b1c89f7765e30bc1c42e4a6da51ca513615660cb8a62ef9a88e3"
 [[package]]
 name = "reth-basic-payload-builder"
 version = "1.6.0"
-source = "git+https://github.com/kevaundray/reth?rev=f17c70c23056cadd7f42ee7c4502f72078d2a118#f17c70c23056cadd7f42ee7c4502f72078d2a118"
+source = "git+https://github.com/kevaundray/reth?rev=0880801d0d6c29d1a7daaee24718b73e5a8a52a8#0880801d0d6c29d1a7daaee24718b73e5a8a52a8"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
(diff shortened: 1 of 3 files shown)


kevaundray:
With this we can close #111

scripts/migrate_mainnet_fixtures_chainconfig.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python3
+"""
+Script to process zkevm fixture JSON files by removing existing chain_config
+and adding a new fixed chain_config inside the block_and_witness field.
+
+This is used to migrate previously generated fixture files to the new format.
```
kevaundray, line 6: Once the migration is done on existing fixtures, do we need this script after that?

kevaundray (approved):

scripts/migrate_mainnet_fixtures_chainconfig.py
jsign: Nop, it shouldn't be needed.
I think my plan is:
- Merge this PR
- Update the witness collector VM we have so it starts to collect using the new format.
- I've to do some work on untar-ing + migrating + re-taring previously collected stuff using this script as part of that process
- After all looks good, I can delete this script in the repo.

jsign: And hopefully we don't touch the witness generator format again :sweat_smile:

PR #143, opened 2025-08-15
ci: use ere images

jsign:
This PR:
- Use `ere` images so we avoid building base and SDK docker images which take a lot of time and generate GH rate limiting problems.
- I switched to using GH runners instead of devops. Now that the pipeline is less heavy it's better, since devops needed also an extra step to instal the C toolchain which takes time. Plus, sometimes it took 2m just to start the workflow compared to GH runners which start immediately.
- Refactor the guest program runs so we avoid duplication

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,8 +108,8 @@ zkevm-metrics = { path = "crates/metrics" }
 benchmark-runner = { path = "crates/benchmark-runner" }
 guest-libs = { path = "ere-guests/libs" }
 
-zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", rev = "cb037368d6ac02486deede9b059202a355b0234f", package = "zkvm-interface" }
-ere-dockerized = { git = "https://github.com/eth-applied-research-group/ere", rev = "cb037368d6ac02486deede9b059202a355b0234f", package = "ere-dockerized" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -108,8 +108,8 @@ zkevm-metrics = { path = "crates/metrics" }
 benchmark-runner = { path = "crates/benchmark-runner" }
 guest-libs = { path = "ere-guests/libs" }
 
-zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", rev = "cb037368d6ac02486deede9b059202a355b0234f", package = "zkvm-interface" }
-ere-dockerized = { git = "https://github.com/eth-applied-research-group/ere", rev = "cb037368d6ac02486deede9b059202a355b0234f", package = "ere-dockerized" }
+zkvm-interface = { git = "https://github.com/eth-applied-research-group/ere", rev = "06d15a4a77feef4772a37bee964dcec5733f9eff", package = "zkvm-interface" }
+ere-dockerized = { git = "https://github.com/eth-applied-research-group/ere", rev = "06d15a4a77feef4772a37bee964dcec5733f9eff", package = "ere-dockerized" }
```
jsign, line 112: I think next week we can cut a release, and change this (and img in workflow) to tags.

PR #145, opened 2025-08-15
Use extracted mpt and interface

jsign:
Testing tentative extracted optimized MPT and interface.

Not meant to be merged.


ere-guests/libs/src/mpt.rs
jsign, line 1: this is gone

Issue #149, opened 2025-08-22
Support executing invalid blocks


jsign:
Support executing/proving invalid blocks. For this we'll need some changes in https://github.com/kevaundray/reth.

This means:
- Supporting generating a partial execution witness up to the point of failure
- Using this partial execution witness to execute/prove it in `stateless-validator` guest program

I'll start working on this now.

@kevaundray, lateral question: do we want to add an extra input to the `stateless-validator` guest program like `must_valid : bool`, or maybe sounds okay that the guest-program panics ([which is what it does today](https://github.com/eth-act/zkevm-benchmark-workload/blob/1755a2d18f5c2ee400589f727de5f59cfb306f70/ere-guests/stateless-validator/sp1/src/main.rs#L36))?

Feels like accepting a new parameter might be polluting a bit the input for a use-case that isn't interesting for mainnet. If we keep letting it panic, we could still assert it in the `zkevm-metrics` since it would be a "crashed" run (i.e., not "successful").

kevaundray:
We could modify the guest program to return a boolean for invalid/valid blocks -- this was related to the usecase we spoke of before -- then we would need to add proof verification and public input checking I think. Then the benchmark software checks if it should have failed against the public input. Does that achieve the same thing?

jsign:
> We could modify the guest program to return a boolean for invalid/valid blocks -- this was related to the usecase we spoke of before -- then we would need to add proof verification and public input checking I think. Then the benchmark software checks if it should have failed against the public input. Does that achieve the same thing?

That sound nice and maybe the ideal endgame. I'm not sure yet if all zkVMs support this to add it as part of the API for ere. cc'ing @han0110 to pick his brain.

jsign:
> you mean proof verification with public inputs? 

Yes

kevaundray:
I think we should probably ping them and ignore the ones that don't have this, since it's somewhat of a base feature

jsign:
To avoid blocking and rushing on the work, small followup task whenever we support checking public outputs https://github.com/eth-act/zkevm-benchmark-workload/issues/151

PR #150, opened 2025-08-25
Support processing invalid blocks

jsign:
This PR adds support for executing/proving invalid blocks.

It relies on a new version of our reth-fork (https://github.com/kevaundray/reth/pull/21).

Close https://github.com/eth-act/zkevm-benchmark-workload/issues/149

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,14 +113,14 @@ ere-dockerized = { git = "https://github.com/eth-applied-research-group/ere", re
 
 # branch is kw/zkevm-benchmark-workload-repo
 # NOTE: We are using a branch of a branch that has not yet been merged into master.
-ef-tests = { git = "https://github.com/kevaundray/reth", rev = "0f3b7fdbdc4c4d9c88607e8f72d651afa2ac3bd4" }
-reth-ethereum-primitives = { git = "https://github.com/kevaundray/reth", rev = "0f3b7fdbdc4c4d9c88607e8f72d651afa2ac3bd4" }
-reth-stateless = { git = "https://github.com/kevaundray/reth", rev = "0f3b7fdbdc4c4d9c88607e8f72d651afa2ac3bd4" }
-reth-rpc-api = { git = "https://github.com/kevaundray/reth", rev = "0f3b7fdbdc4c4d9c88607e8f72d651afa2ac3bd4" }
-reth-primitives-traits = { git = "https://github.com/kevaundray/reth", rev = "0f3b7fdbdc4c4d9c88607e8f72d651afa2ac3bd4" }
-reth-errors = { git = "https://github.com/kevaundray/reth", rev = "0f3b7fdbdc4c4d9c88607e8f72d651afa2ac3bd4" }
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -42,12 +42,12 @@ jobs:
           fi
 
   stateless-validator:
```
jsign, line 44: I think this is already getting a bit complex, so I'm thinking on moving integration testing as Rust tests in a separate crate. 

Also because most of this integration testing would be also useful to run locally with a single command or similar, and not rely on CI.

crates/witness-generator/src/eest_generator.rs
```diff
@@ -385,4 +381,40 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_invalid_block() -> Result<()> {
```
jsign, line 386: An extra test to dig more into the details on the partial witness from an invalid block. For this case, we should see the witness to be filled but only with the parent header data since the block is invalid due to parent block checks. The rest of the witness fields must be empty.

crates/witness-generator/src/eest_generator.rs
```diff
@@ -170,7 +166,7 @@ impl WitnessGenerator for ExecSpecTestBlocksAndWitnesses {
                 let chain_config = chain_spec.genesis.config;
                 Ok(BlockAndWitness {
                     name: name.to_string(),
-                    block_and_witness: run_case(case)?
+                    block_and_witness: BlockchainTestCase::run_single_case(name, case)?
```
jsign, line 169: We switch to `BlockchainTestCase::run_single_case`. This method (https://github.com/eth-act/zkevm-benchmark-workload/pull/150) semantics is that it only returns an error if the test fails. 

Note that if the test has invalid blocks and Reth failed (as expected), that means the test run is successful.

This means that from our perspective here, we always expect all fixtures to succeed (we don't care if they have valid or invalid blocks). The returned witness is always everything needed to execute the block (i.e., full or partial).

.github/workflows/integration-tests.yml
```diff
@@ -118,9 +123,18 @@ jobs:
...
           
-          if grep -r "crashed" "$zkvm_folder"/*.json 2>/dev/null; then
-            echo "Found 'crashed' string in ZKVM folder JSON files"
-            exit 1
+          if [ "${{ matrix.test-type }}" = "invalid-blocks" ]; then
+            # For invalid-blocks, we expect all tests to crash - fail if any success is found
+            if grep -r "success" "$zkvm_folder"/*.json 2>/dev/null; then
+              echo "Found unexpected 'success' string in invalid-blocks test - some tests succeeded when they should have failed"
+              exit 1
+            fi
```
jsign, line 131: To avoid blocking this PR on ere supporting reporting outputs, for now I check that the run crashed (since the guest program today panics if the stateless validation of the block fails).

Whenever ere supports reporting outputs, we can easily check outputs instead of doing this. (sidenote: since I'll be moving integration tests into Rust code, it will also be nicier than looking from strings in json files which I think is messy).

kevaundray (approved):
changes and followup issues look good to me

PR #152, opened 2025-08-26
Run integration test cases in dedicated crate

jsign:
The current way integration tests run in CI is through a GH workflow which is starting to become quite overloaded with logic. Plus, it wasn't possible to easily run all the same tests locally.

This Pr changes that to have a new `integration-tests` crate which has the same coverage as before, which allows running locally and the CI to become simpler.


.github/workflows/integration-tests.yml
```diff
@@ -41,51 +41,48 @@ jobs:
...
-        run: |
-          tar -xzvf tests/mainnet-zkevm-fixtures-input.tar.gz
-          mv mainnet-zkevm-fixtures-input zkevm-fixtures-input
-      
-      - name: Prepare invalid blocks fixtures
-        if: matrix.test-type == 'invalid-blocks'
-        run: |
-          RUST_LOG=info cargo run -p witness-generator-cli --release -- tests --eest-fixtures-path tests/eest-invalid-block-fixtures 
-
       - name: Pull ere images
```
jsign, line 86: ere images are still pulled to save time in the CI.

.github/workflows/integration-tests.yml
```diff
@@ -96,121 +93,4 @@ jobs:
...
-          fi
-
-      - name: Upload metrics artifact
-        if: always()
-        uses: actions/upload-artifact@v4
-        with:
-          name: zkevm-metrics-empty-program-${{ matrix.zkvm }}-${{ matrix.action }}
-          path: zkevm-metrics
-          retention-days: 14
+        run: CI=1 RUST_LOG=info,sp1_core_executor=warn ZKVM=${{ matrix.zkvm }} cargo test --release -p integration-tests -- --test-threads=1 ${{ matrix.test }}
```
jsign, line 96: TL;DR here:
- `CI=1` is to pipe the expected env that ZisK expects when you only use CPU as a proving resource. If we don't do this, the build-args change and it tries to rebuild the pulled ere image to compile NVIDIA libraries which we don't use in CI (thus why the env name).
- `ZKVM` is an env var in the new integration test crate which allows to filter which zkvms are run. If you run locally without setting this env-var, it will run for all zkvms that make sense. The logic of using this env var is to run stuff more in parallel only in CI, and also identify quicker which zkvm might be failing (since each one here is a matrix combination)
- `--test-threads=1` is to avoid clashes because of zkvm patches.

.github/workflows/rust-checks.yml
```diff
@@ -39,8 +39,6 @@ jobs:
       
       - name: Cache dependencies
         uses: Swatinem/rust-cache@v2
-        with:
-          key: base
```
jsign, line 43: Removed since before we had many `--features` -- now that we use dockerized ere, we can use a single key for Rust cache.

crates/benchmark-runner/src/lib.rs
```diff
@@ -136,3 +138,52 @@ fn get_panic_msg(panic_info: Box<dyn Any + Send>) -> String {
         .or_else(|| panic_info.downcast_ref::<String>().cloned())
         .unwrap_or_else(|| "Unknown panic occurred".to_string())
 }
+
+/// Creates the requested zkVMs configured for the guest program and resources.
```
jsign, line 142: Just moved from `ere-hosts/main.rs` since now is useful for integration tests.

tests/assets/eest-empty-block/fixtures/blockchain_tests/test_worst_compute.py::test_empty_block[fork_Prague-benchmark-gas-value_1M-blockchain_test].json
Comment (jsign) [inline tests/assets/eest-empty-block/fixtures/blockchain_tests/test_worst_compute.py::test_empty_block[fork_Prague-benchmark-gas-value_1M-blockchain_test].json:1]: As a general rule, I'm targeting to always have assets in the repo -- so we can run the tests without internet access. 

You might still need internet for building docker images (or pulling ere ones), but if you have them cached isn't a problem.

tests/src/empty_program.rs
```diff
@@ -0,0 +1,38 @@
+#[cfg(test)]
```
jsign, line 1: There's one module per guest program in the integration test.

tests/src/empty_program.rs
```diff
@@ -0,0 +1,38 @@
...
+        assert_executions_successful, assert_proving_successful, get_env_zkvm_or_default, run_guest,
+    };
+
+    use benchmark_runner::{guest_programs, Action};
+    use ere_dockerized::ErezkVM;
+    use tempfile::tempdir;
+
+    #[tokio::test]
+    async fn execute_empty_program() {
+        let zkvms = get_env_zkvm_or_default(vec![ErezkVM::SP1, ErezkVM::Risc0, ErezkVM::Zisk]);
```
jsign, line 13: This is how we get the zkvm as targets for the run. `get_env_zkvm_or_default` is a helper that reads the `ZKVM` env-var (if any) or fallbacks to the provided default.

tests/src/stateless_validator.rs
jsign, line 1: Nothing super interesting here -- mostly relying on stacked helpers to express test intentions.

tests/src/stateless_validator.rs
```diff
@@ -0,0 +1,114 @@
+#[cfg(test)]
```
jsign, line 1: Nothing super interesting here -- mostly relying on stacked helpers to express test intentions.

PR #155, opened 2025-08-29
ere-guests: commit to public inputs for SP1 and Risc0

### LINKED ISSUE #151: Avoid panicking on `stateless-validator` for invalid blocks
Whenever `ere` supports public outputs, change `stateless-validator` guest to return if the block execution was successful instead of panicking.

This also means that in CI we can change block failure detection without checking for run crashes. https://github.com/eth-act/zkevm-benchmark-workload/pull/150#discussion_r2298596396

jsign:
This PR:
- Updates to ere, which now has ample support for checking public input
- Refactor internals to introduce the concept of `GuestIO` which now contemplates inputs and _expected_ outputs.
- Implement public input commitments for SP1 and Risc0 -- if things feel okay, we can do for the rest of guest programs next or when they're ready to be benchmarked, since most can't be run yet.

This is a WIP, so it is not ready for a full review.

Closes #151


crates/benchmark-runner/src/runner.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// Generate a proof for the zkVM execution
+    Prove,
+    /// Only execute the zkVM without proving
+    Execute,
+}
+
+/// Executes benchmarks for a given guest program type and zkVM
+pub fn run_benchmark<M, OV>(
+    zkvm: ErezkVM,
+    ere_zkvm: &EreDockerizedzkVM,
```
jsign, line 39: @han0110, I think it would be better to have this [zkvm()](https://github.com/eth-act/ere/blob/664dc62c9fa51ba3ce50b40c474644c723228819/crates/ere-dockerized/src/lib.rs#L220-L222) in `EreDockerizedzkVM` instead of `EreDockerizedCompiler`.
If that's the case, I can remove this first argument which feels a bit duplicate.

I'm still wondering if there's a better reason why you thought of putting it in the compiler trait instead. Up to now in my implementation, I haven't used it.

---
The PR is still not ready for review (have to finish some uninteresting stuff), but if you're curious on the gist:
- See how things are commited in [guest program](https://github.com/eth-act/zkevm-benchmark-workload/pull/155/files#diff-8ea2b563dd2f17483b54a0eba81176cce8b5716a8f39051a3b33cbfe24bdaf65R44-R61).
- And how [they're checked](https://github.com/eth-act/zkevm-benchmark-workload/pull/155/files#diff-3ed7ecd0adbc670eb64fbdafda3ffb3e48244dc89548d29e64dfad6a60d47017R79-R115)
---

han0110: Ah you are right, I was intended to add the `zkvm` method to `EreDockerizedzkVM` instead of `EreDockerizedCompiler`! Will update it in a PR soon.

---

The public values committed and checking looks nice!

PR #171, opened 2025-09-09
Add Pico to integration tests

### LINKED ISSUE #43: feat: Fix Pico integration
Currently the Pico code has been commented out due to #42

jsign:
This PR adds Pico to integration tests:
- [x] Do corresponding fixes `stateless-validator`
- [x] Create `empty-program` and `panic-guest` guests
- [x] Double-check precompiles are up to date
- [x] Enable in `tests` integration test package targeting all possible coverage
- [x] Enable in CI

Discovered some issues/blockers:
- #172 
  - #173
  - #175
- #174 

Fixes #43


precompile-patches/pico.toml
```diff
@@ -1,16 +1,7 @@
 [patch.crates-io]
```
jsign, line 1: This was quite messy to figure out. I removed everything that generated a warning, always trying to double check if there's a correct patch to be applied.

I also asked Brevis people if they've a place to double check what they expect should apply. I haven't got an answer yet, but if they provide something to double check and there's something missing, I'll add in further PR.

tests/src/stateless_validator.rs
```diff
@@ -18,41 +18,77 @@ mod tests {
...
+            (ExecutionClient::Ethrex, ErezkVM::SP1),
+            // (ExecutionClient::Ethrex, ErezkVM::Risc0), // See https://github.com/eth-act/ere/issues/121
+            // (ExecutionClient::Ethrex, ErezkVM::OpenVM), // See https://github.com/eth-act/ere/issues/168
+            // (ExecutionClient::Ethrex, ErezkVM::Pico), // See https://github.com/eth-act/ere/issues/174
+        ]);
+        empty_block(Action::Execute, &el_zkvms).await;
     }
 
     #[tokio::test]
     async fn execute_mainnet_blocks() {
```
jsign, line 50: Instaed of doing a single `run_guest` with all the zkvms, now we iterate on each and run independently.

This is requried since now Pico needs to do some exceptions to avoid running specific tests that we must run in others... so the json files aren't the same set for all zkvms anymore (thus why we run independently now).

PR #176, opened 2025-09-09
ethrex: update with removal of trust assumptions

### LINKED ISSUE #165: ethrex: remove input trust assumptions
We mentioned the Ethrex team that their `ProgramInput` has hashed RLP nodes and bytecodes. While this helps in performance since it avoids hashing on the guest side, this is a trust assumption that shouldn't exist.

This is currently addressed in https://github.com/lambdaclass/ethrex/pull/4337.

When this PR is merged, update ethrex in this repo and adjust accordingly the input construction.

jsign:
Update Ethrex which now has the just merged https://github.com/lambdaclass/ethrex/pull/4337

Fixes #165

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -121,9 +121,9 @@ reth-errors = { git = "https://github.com/kevaundray/reth", rev = "fa3efe112aaaf
 reth-trie-common = { git = "https://github.com/kevaundray/reth", rev = "fa3efe112aaaf1d18b56212d537a83c3cc681782" }
 reth-chainspec = { git = "https://github.com/kevaundray/reth", rev = "fa3efe112aaaf1d18b56212d537a83c3cc681782" }
 
-ethrex-guest-program = { git = "https://github.com/lambdaclass/ethrex.git", rev = "c8ce1f272241106f4bdbfc5028c0526f75189e1e", package = "guest_program" }
-ethrex-common = { git = "https://github.com/lambdaclass/ethrex.git", rev = "c8ce1f272241106f4bdbfc5028c0526f75189e1e", default-features = false }
-ethrex-rlp = { git = "https://github.com/lambdaclass/ethrex.git", rev = "c8ce1f272241106f4bdbfc5028c0526f75189e1e", default-features = false }
+ethrex-guest-program = { git = "https://github.com/lambdaclass/ethrex.git", rev = "5b4c943007b96152453e5a8f8a1a592608b11ea0", package = "guest_program" }
+ethrex-common = { git = "https://github.com/lambdaclass/ethrex.git", rev = "5b4c943007b96152453e5a8f8a1a592608b11ea0", default-features = false }
+ethrex-rlp = { git = "https://github.com/lambdaclass/ethrex.git", rev = "5b4c943007b96152453e5a8f8a1a592608b11ea0", default-features = false }
 
 
 # alloy
--- a/crates/benchmark-runner/src/stateless_validator.rs
+++ b/crates/benchmark-runner/src/stateless_validator.rs
@@ -1,19 +1,17 @@
 //! Stateless validator guest program.
 
-use std::{collections::BTreeMap, convert::TryInto, path::Path};
+use std::{convert::TryInto, path::Path};
 
 use alloy_eips::eip6110::MAINNET_DEPOSIT_CONTRACT_ADDRESS;
-use alloy_primitives::keccak256;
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -121,9 +121,9 @@ reth-errors = { git = "https://github.com/kevaundray/reth", rev = "fa3efe112aaaf
 reth-trie-common = { git = "https://github.com/kevaundray/reth", rev = "fa3efe112aaaf1d18b56212d537a83c3cc681782" }
 reth-chainspec = { git = "https://github.com/kevaundray/reth", rev = "fa3efe112aaaf1d18b56212d537a83c3cc681782" }
 
-ethrex-guest-program = { git = "https://github.com/lambdaclass/ethrex.git", rev = "c8ce1f272241106f4bdbfc5028c0526f75189e1e", package = "guest_program" }
-ethrex-common = { git = "https://github.com/lambdaclass/ethrex.git", rev = "c8ce1f272241106f4bdbfc5028c0526f75189e1e", default-features = false }
-ethrex-rlp = { git = "https://github.com/lambdaclass/ethrex.git", rev = "c8ce1f272241106f4bdbfc5028c0526f75189e1e", default-features = false }
+ethrex-guest-program = { git = "https://github.com/lambdaclass/ethrex.git", rev = "5b4c943007b96152453e5a8f8a1a592608b11ea0", package = "guest_program" }
```
jsign, line 124: https://github.com/lambdaclass/ethrex/commit/5b4c943007b96152453e5a8f8a1a592608b11ea0

crates/benchmark-runner/src/stateless_validator.rs
```diff
@@ -181,26 +179,9 @@ fn write_stdin(si: &StatelessInput, el: &ExecutionClient) -> Result<Input> {
...
-        .iter()
-        .map(|h| Ok(BlockHeader::decode(h.as_ref())?))
-        .map(|h| h.map(|h| (h.number, h)))
-        .collect::<Result<BTreeMap<u64, BlockHeader>>>()?;
-
-    let parent_block_header = block_headers
-        .get(&(block_number - 1))
-        .ok_or_else(|| anyhow::anyhow!("Missing parent block header"))?
-        .clone();
+) -> Result<ExecutionWitness> {
```
jsign, line 182: This got quite simplified since we don't need to calculate BTreeMaps

ere-guests/stateless-validator/ethrex/sp1/src/main.rs
```diff
@@ -8,11 +8,14 @@ sp1_zkvm::entrypoint!(main);
...
     let input = sp1_zkvm::io::read_vec();
-    let input = rkyv::from_bytes::<ProgramInput, Error>(&input).unwrap();
+    let mut input = rkyv::from_bytes::<ProgramInput, Error>(&input).unwrap();
     println!("cycle-tracker-report-end: read_input");
 
     println!("cycle-tracker-report-start: public_inputs_preparation");
     let block_hash = input.blocks[0].hash();
+    // Ethrex forces the block hash of any provided block to be zero, instead of accepting pre-calculated values
+    // and just checking them again.
+    input.blocks[0].header.hash = Default::default();
```
jsign, line 18: This is a bit unfortunate since we're discarding a keccak calculation.

The reason for this is [here](https://github.com/lambdaclass/ethrex/blob/d75dc988eb75b31b35a8ef439f60d07bf914e362/crates/common/types/block_execution_witness.rs#L498).

The received `blocks` must not have their `hash` already calculated. I suspect this is because they want to be strict on not received precomputed hashes as inputs.

For anothe angle, here in L15 we do `let block_hash = input.blocks[0].hash();` which indirectly caches the calculated hash, thus the provided `block` does not meet the expectation later inside the block validation logic (linked above).

I think ideally Ethrex should accept the hashes value to be set, but always compute it in the validation and compare it (if present). This would make things a bit more flexible here, while keeping the "do not trust" values.

Note that the reason why I want the hash is to do the L23 check, so in the case of wanting to prove more than one block I can still commit to the block hash. I could simply avoid that expectation and let the host see that the third output (i.e., successful block validation) was `false` anyway, but feels like an unnecessary compromise to workaround this.

I also thought about doing the `.len() != 1` after the `execution_program`, but that's a bit messy since `res` is a `Result`, plus I can't do `input.blocks[0]` since `input` was moved in `execution_program(input)` call.

PR #178, opened 2025-09-10
ZisK guest reveal sha256 digested pubilc inputs

han0110:
- In ZisK reth guest reveal sha256 digested public inputs
- Check the public inputs in `benchmark-runner` for ZisK using the same logic as for OpenVM
- Add `ziskos` patch to make sure all `ziskos` dependency has the same version `v0.11.0`, to avoid duplicate exported symbols (some of the patch doesn't specify the tag and cargo recognize it to be different version)
- Add partial patches for ZisK (other patches couldn't be used because they depend on `sp1-lib` v4 which creates conflict)


ere-guests/empty-program/zisk/Cargo.toml
```diff
@@ -6,7 +6,7 @@ rust-version.workspace = true
 license.workspace = true
 
 [dependencies]
-ziskos = { git = "https://github.com/0xPolygonHermez/zisk.git", rev = "f9a3655" }
+ziskos = { git = "https://github.com/0xPolygonHermez/zisk.git" }
```
han0110, line 9: This will be patched to `v0.11.0` as well

precompile-patches/zisk.toml
```diff
@@ -1,2 +1,6 @@
+# Copied from https://github.com/0xPolygonHermez/zisk-eth-client/blob/main/bin/client/Cargo.toml
 [patch.crates-io]
-tiny-keccak = { git = "https://github.com/0xPolygonHermez/zisk-patch-tiny-keccak.git", branch = "zisk" }
+sha2 = { git = "https://github.com/0xPolygonHermez/zisk-patch-hashes.git", branch = "zisk-patch-sha2/v0.10.9" }
+sha3 = { git = "https://github.com/0xPolygonHermez/zisk-patch-hashes.git", branch = "zisk-patch-sha3/v0.10.8" }
+k256 = { git = "https://github.com/0xPolygonHermez/zisk-patch-elliptic-curves.git", branch = "zisk-patch-k256/v0.13.4" }
+bn = { git = "https://github.com/0xPolygonHermez/zisk-patch-bn.git", branch = "zisk-patch-bn/v0.6.0", package = "substrate-bn" }
```
han0110, line 6: We skip the `sp1_bls12_381` patching because we can turn off the `kzg-rs` feature. The patching `sp1_bls12_381` seems only to disable the `kzg-rs` feature that being turned on in `rsp`.

ere-guests/Cargo.toml
```diff
@@ -161,4 +162,8 @@ ethereum_ssz = "0.9"
 
 risc0-ethereum-trie = { git = "https://github.com/risc0/risc0-ethereum.git", rev = "4a5efe13b3655258b89e7082c57743e445133dc8" }
 
+# Patch `ziskos` of `zisk-patch-*` to pin to the same version of guest.
+[patch."https://github.com/0xPolygonHermez/zisk.git"]
+ziskos = { git = "https://github.com/0xPolygonHermez//zisk.git", tag = "v0.11.0" }
```
han0110, line 167: Patch `ziskos` to pin the patches' dependency version to the same one of the guest ([the patches don't specify any version](https://github.com/0xPolygonHermez/zisk-patch-hashes/blob/zisk-patch-sha2/v0.10.9/sha2/Cargo.toml#L26) and cargo will recognize them as different crates).

ere-guests/empty-program/zisk/Cargo.toml
jsign: If ZisK default branch one day moves to v0.12.0 and Cargo.lock is updated, then the patch in the other file won't be applied anymore, no?
Wouldn't be safer to keep the `tag="v0.11.0"` here?

ere-guests/Cargo.toml
jsign: Uhm, never saw a patch that changes a source to the same one.

Say that a zisk patch ends up resolving to v0.12.0 in its cargo lock -- I don't think this patch will force them to use ziskos v0.11.0, no?

The way I read patches is usually to replace one source with another for a specific semver. I don't think they can pin versions of dependencies, only replace from where they're resolved if are part of the dependency graph.

(I'm not an expert of cargo patches, so sharing my thinking here to double check)

Nit: double `/` in url.

han0110:
> If ZisK default branch one day moves to v0.12.0 and Cargo.lock is updated, then the patch in the other file won't be applied anymore, no?
Wouldn't be safer to keep the `tag="v0.11.0"` here?

For crate the use `git = "{url}"` we can patch to arbitrary version, so I think even the default branch gets update, the patch will still take effect, so I think it'd be better to not specify the tag in many places (I was afraid one could modify the dependency tag but not the patch version, and in the end it's still using the old version).

> Nit: double / in url, but I can fix it in upcoming pr.

Actually the double / is intended, it's used to trick cargo to recognize it as a different source (tho they resolve to the same one), if we remove it the patch will be invalid.

---

I think the main problem is still those crate `zisk-patch-*` not specifying the tag, so we have to patch them to pin them to the same version the ere uses (otherwise the host/guest could be incompatible).

jsign:
> Actually the double / is intended, it's used to trick cargo to recognize it as a different source (tho they resolve to the same one), if we remove it the patch will be invalid.

No way :smile: 

@han0110, could we ask ZisK guys to use tagged version in patches? I think its an easy ask and we could remove this opaque hack.

PR #180, opened 2025-09-10
pico: add kzg-rs patch which fixes skipped integration test mainnet blocks

### LINKED ISSUE #175: Pico: executing blocks 22974584 and 22974587 fails validation
In the `execute_mainnet_blocks` integration tests we run 15 consecutive mainnet blocks.

13 of them execute fine (i.e., emulation + public output verification).
2 of them [(22974584 and 22974587) fail to execute](https://github.com/eth-act/zkevm-benchmark-workload/blob/f04f05a2be4fd45165a33e6f5c42d829f7e79ce5/tests/src/stateless_validator.rs#L72).

"Fail to execute" means that the expected output for `success` isn't `true` but `false`, which indicates that Reth failed to validate this block.

This sounds quite bad, since all the blocks should validate correctly (i.e., the rest of zkVMs work fine with Reth, so it is a Pico issue).

We need to dig a bit more to understand what might be happening here, since this isn't a crash but an incorrect execution which sound suprising.

jsign:
Fixes #175

### DIFF
--- a/precompile-patches/pico.toml
+++ b/precompile-patches/pico.toml
@@ -3,6 +3,6 @@ tiny-keccak = { git = "https://github.com/brevis-network/tiny-keccak", tag = "pi
 sha2 = { git = "https://github.com/brevis-network/hashes", rev = "2d0f0f9f4abde803c07801ca2fa02fd7911c6b30", package = "sha2" }
 sha3 = { git = "https://github.com/brevis-network/hashes", tag = "pico-patch-v1.0.1-sha3-v0.10.8", package = "sha3" }
 curve25519-dalek = { git = "https://github.com/brevis-network/curve25519-dalek", tag = "pico-patch-v1.0.1-curve25519-dalek-v4.1.3" }
-kzg-rs = { git = "https://github.com/brevis-network/kzg-rs", rev = "7f034a1a0f9328e7c2ae667a5b6cb98bf6ab2a79" }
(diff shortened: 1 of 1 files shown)


ere-guests/stateless-validator/reth/pico/Cargo.toml
```diff
@@ -19,7 +19,10 @@ reth-primitives-traits = { workspace = true, features = [
 reth-evm-ethereum = { workspace = true }
 reth-chainspec = { workspace = true }
 
-revm = { workspace = true, default-features = false, features = ["bn"] }
+revm = { workspace = true, default-features = false, features = [
+    "kzg-rs",
```
jsign, line 23: The cause of block failure was a runtime failure of revm detecting not `c-kzg` nor `kzg-rs` were enabled. The reason why `kzg-rs` wasn't enabled was because of this Pico [compilation error](https://pastebin.com/raw/JjCLW51w) I found yesterday. This was confirmed by the Brevis team.

But, I tried an unodcumented precompile patch of `kzg-rs` directly that was merged [some hours ago](https://github.com/brevis-network/kzg-rs/commit/7f034a1a0f9328e7c2ae667a5b6cb98bf6ab2a79), and this fixes the compilation problem thus the runtime problem too.

jsign: That branch doesn't compile

PR #187, opened 2025-09-17
chore: update ere to v0.0.12

jsign:
Update ere to v0.0.12. In particular, because it has a temporary workaround to allow using ZisK without patching ere's dep in this repo (https://github.com/eth-act/ere/pull/137).

To use ZisK remember to export:
- `CUDA_ARCH=sm_{xx}` to your corresponding GPU arch.
- For 4090/L40s you will need also `ZISK_UNLOCK_MAPPED_MEMORY=1`

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -12,7 +12,7 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
-  ERE_TAG: 0.0.12-2f979db
+  ERE_TAG: 0.0.12-7ef4598
   OPENVM_RUST_TOOLCHAIN: nightly-2025-08-07
 
 jobs: 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1548,7 +1548,7 @@ dependencies = [
 [[package]]
 name = "build-utils"
 version = "0.0.12"
-source = "git+https://github.com/eth-applied-research-group/ere?rev=2f979dbd01fc11a08a6c9d650f5f8fad698e8469#2f979dbd01fc11a08a6c9d650f5f8fad698e8469"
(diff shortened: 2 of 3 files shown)


jsign:
`ere-cli-risc0` image is [still running for many hours in ere](https://github.com/eth-act/ere/actions/runs/17797158390/job/50587416973), prob this CI will fail. Whenever ere finishes, I'll re-run.

jsign:
Now I'm wondering about something reg ZisK not pinning versioning could be related.

jsign:
Yeah, as expected removing the precompile works fine. I'll create a PR later.
Good that we ended up updating our Reth fork to the latest that had the CPU fallback for the kzg lib dependency:sweat_smile:

jsign:
Yeah, I think if you revert it to the previous version, we can keep things as they are.

I think we could still remove this patched precompile, but feels we're indirectly also making a decision on the "official" way ZisK wants to resolve KZG verification. If we remove it, we're kind of responsible of saying arkworks lib is the right decision (since IIRC is waht Reth uses now as the fallback).

jsign:
Uhm, I don't find convincing we need to maintain this patch long-term.

If we see this is the case, maybe we can think of moving zisk out of the ere-guest workspace to avoid the conflict. It might not be nice from the workload side to have this exception, but from another angle I would prefer us not having to maintain patches long term (e.g. temporarily waiting for a next release is fine, but long term I don't like it that much).

I guess we can wait until their next v0.12.0 release, see how things work there, and we can reveisit what to do.

han0110:
yes I just reverted it to old version, tho the git revision might be different.

jsign:
> yes I just reverted it to old version, tho the git revision might be different.

Re-tested with a reset Cargo.lock and the patch activated again, all good :+1:

PR #190, opened 2025-09-24
Automatic website gen for integration test cases executions

jsign:
This PR's primary goal is to leverage the work done in the integration test pipeline to generate an automated website displaying zkVM execution results.

Summary of changes:
- The integration tests allow a new optional `WORKLOAD_OUTPUT_DIR` so the test results aren't discarded, but saved to a desired path. The way it works is that it uses `tempdir` (as it always did yesterday) or a defined path depending on that var.
- In CI, we use this new `WORKLOAD_OUTPUT_DIR` so all the integration test results runs are uploaded as artifacts. This can be useful to further inspect runs, or have a medium-term historical view if useful.
- After integration tests are run, a new GH job collects all uploaded artifacts, it merges them (since integration tests are run in parallel for different EL/zkVM combinations), and generates a website using a Python script. Then, another job after deploys this website.
- For the `stateless-validator` guest program runs, the `zkevm-metrics` folder structure has a new level "EL". So instead of now being `zkevm-metrics/<zkvm>/[results]` is `zkevm-metrics/<el>/<zkvm>/[results]`. Apart for being more organized, it was useful to allow the generated website to identify ELs correctly.
- I reorganized the integration test jobs extracting runs in a separate component, which avoids some duplication that was present and potential future reuse.

The idea is to try to generate execution reports automatically:
- on every merge to master, since we can get that somewhat for free, since we were already running integration tests. Today this covers 30 mainnet blocks, invalid block, and empty block, but anything else we add in the future will get included easily.
- I'll experiment with whether we have a separate daily runner for more EEST cases. I'm not sure the CI machines can run all of them. If it is too heavy, I could manually choose the already identified worst cases to at least keep track of those most important ones and see which zkVM crash or are too slow (which is actually the most important signal we want to track). 

https://eth-act.github.io/zkevm-benchmark-workload/


.github/workflows/integration-tests.yml
```diff
@@ -10,12 +10,15 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
jsign, line 10: After the integration test refactoring, as a side effect, the CI runs are a bit better organized in the UI:
image

.github/workflows/integration-tests.yml
```diff
@@ -55,13 +57,17 @@ jobs:
           - prove_empty_block
         zkvm: [sp1, risc0, pico, zisk, openvm]
         el: [reth, ethrex]
+        include:
+          - zkvm: openvm
+            threads: 2
```
jsign, line 62: This is an old setup since OpenVM uses quite a lot of RAM per case, so going very far in parallelization makes the CI machine struggle. But since OpenVM emulation is very fast, doesn't hurt much CI duration.

For the rest, I configured 12 threads to avoid going with 16 which is the default for the CI machine. Mainly to give some stability to wall-clock time runs.

.github/workflows/integration-tests.yml
```diff
@@ -74,37 +80,17 @@ jobs:
...
-            dst="ere-${variant}:${ERE_TAG}"
-            docker pull "$src"
-            docker tag "$src" "$dst"
-          done
-
-      - name: Run benchmark
-        run: |
-          ${{ matrix.zkvm == 'openvm' && 'RAYON_NUM_THREADS=1' || '' }} RUST_LOG=warn,benchmark_runner=info ZKVM=${{ matrix.zkvm }} EL=${{ matrix.el }} cargo test --release -p integration-tests -- --test-threads=1 ${{ matrix.test }}
+            zkvm: zisk # See https://github.com/eth-act/ere/issues/186
+    uses: ./.github/workflows/run-benchmark.yml
```
jsign, line 84: Most of deleted code in this file was extracted to this separate component for better modularity and avoid repetition.

.github/workflows/integration-tests.yml
```diff
@@ -120,27 +106,95 @@ jobs:
...
           - test: prove_panic_guest
             zkvm: zisk
+    uses: ./.github/workflows/run-benchmark.yml
+    with:
+      test: ${{ matrix.test }}
+      zkvm: ${{ matrix.zkvm }}
+      el: 'none'
+      threads: 12
 
+  generate-benchmark-website:
```
jsign, line 116: This and next jobs are new ones pulling uploaded artifacts, merging, generating website and publishing.

Pending to only run this on `master` -- for now is always running to test the PR.

crates/ere-hosts/src/main.rs
```diff
@@ -185,6 +178,12 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                 resource,
                 apply_patches,
             )?;
+            let config = RunConfig {
+                output_folder: cli.output_folder,
+                sub_folder: Some(el.as_ref().to_lowercase()),
```
jsign, line 183: I would like to eventually reconsider this solution, but probably closer when we accept raw-ELFs potentially for externally compiled ELFs.

.github/workflows/integration-tests.yml
jsign: Pipeline: 
image

PR #200, opened 2025-10-07
guests: add extra public inputs to do proper CL checks

jsign:
This is a workaround until we decide on a full solution to address https://github.com/ethereum/consensus-specs/issues/4637. 

This PR adds three extra public inputs to do proper assertions in zkCLs regarding the EL block validity.

### DIFF
--- a/crates/benchmark-runner/src/stateless_validator.rs
+++ b/crates/benchmark-runner/src/stateless_validator.rs
@@ -15,6 +15,7 @@ use ethrex_common::{
 };
 use ethrex_guest_program::input::ProgramInput;
 use ethrex_rlp::decode::RLPDecode;
+use guest_libs::blobs::calculate_versioned_hashes_hash;
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use reth_stateless::StatelessInput;
 use rkyv::rancor::Error;
@@ -61,6 +62,16 @@ pub fn stateless_validator_inputs(
                 output: ProgramOutputVerifier {
                     block_hash: bw.block_and_witness.block.hash_slow().0,
                     parent_hash: bw.block_and_witness.block.parent_hash.0,
+                    versioned_hashes_hash: calculate_versioned_hashes_hash(
+                        &bw.block_and_witness.chain_config,
+                        &bw.block_and_witness.block,
+                    ),
+                    parent_beacon_block_root: bw
+                        .block_and_witness
+                        .block
+                        .parent_beacon_block_root
+                        .map(|root| root.0),
+                    requests_hash: bw.block_and_witness.block.requests_hash.map(|h| h.0),
                     success: bw.success,
                 },
             })
@@ -96,6 +107,9 @@ pub fn read_benchmark_fixtures_folder(path: &Path) -> Result<Vec<BlockAndWitness
 pub struct ProgramOutputVerifier {
     block_hash: [u8; 32],
     parent_hash: [u8; 32],
+    versioned_hashes_hash: Option<[u8; 32]>,
+    parent_beacon_block_root: Option<[u8; 32]>,
+    requests_hash: Option<[u8; 32]>,
     success: bool,
 }
 
@@ -106,6 +120,10 @@ impl OutputVerifier for ProgramOutputVerifier {
                 let mut bytes: &[u8] = bytes;
                 let block_hash: [u8; 32] = zkvm.deserialize_from(&mut bytes)?;
                 let parent_hash: [u8; 32] = zkvm.deserialize_from(&mut bytes)?;
+                let versioned_hashes_hash: Option<[u8; 32]> = zkvm.deserialize_from(&mut bytes)?;
+                let parent_beacon_block_root: Option<[u8; 32]> =
+                    zkvm.deserialize_from(&mut bytes)?;
+                let requests_hash: Option<[u8; 32]> = zkvm.deserialize_from(&mut bytes)?;
                 let success: bool = zkvm.deserialize_from(&mut bytes)?;
 
                 if block_hash != self.block_hash {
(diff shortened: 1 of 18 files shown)


crates/benchmark-runner/src/stateless_validator.rs
```diff
@@ -61,6 +62,16 @@ pub fn stateless_validator_inputs(
...
+                    versioned_hashes_hash: calculate_versioned_hashes_hash(
+                        &bw.block_and_witness.chain_config,
+                        &bw.block_and_witness.block,
+                    ),
+                    parent_beacon_block_root: bw
+                        .block_and_witness
+                        .block
+                        .parent_beacon_block_root
+                        .map(|root| root.0),
+                    requests_hash: bw.block_and_witness.block.requests_hash.map(|h| h.0),
```
jsign, line 74: As a reminder `ProgramOutputVerifier` contains the expected values for public input checking. Here, we add the values expected for the new inputs.

This would be akin to what a zkCL verifier would have to do to pass the corresponding public inputs.

ere-guests/libs/src/blobs.rs
```diff
@@ -0,0 +1,27 @@
...
+        .then_some(
+            Sha256::digest(block.body.blob_versioned_hashes_iter().fold(
+                Vec::new(),
+                |mut acc, hash| {
+                    acc.extend_from_slice(&hash.0);
+                    acc
+                },
+            ))
+            .into(),
+        )
```
jsign, line 26: TL;DR:
- If `Cancun` isn't activated, the public input is `None`.
- If it is, then we do `sha256(tx1.blob_hashes || tx2.blob_hashes || ...)`.

The reason why I chose `sha256` instead of `keccak` is to be aligned with the same rationale used in [EIP-7685 for request hashes](https://eips.ethereum.org/EIPS/eip-7685#block-header)... which I think make this decision since `sha256` is friendly for CL compared to `keccak` which is aligned what we are trying to do there too. (Until we have an endgame solution, and prob remove this module)

ere-guests/stateless-validator/ethrex/risc0/src/main.rs
```diff
@@ -18,6 +19,26 @@ fn main() {
     // and just checking them again.
     input.blocks[0].header.hash = Default::default();
     let parent_hash = input.blocks[0].header.parent_hash;
+    let versioned_hashes_hash: Option<[u8; 32]> = input
```
jsign, line 22: Note this is for Ethrex.

ere-guests/stateless-validator/ethrex/sp1/src/main.rs
```diff
@@ -17,12 +18,35 @@ pub fn main() {
     // and just checking them again.
     input.blocks[0].header.hash = Default::default();
     let parent_hash = input.blocks[0].header.parent_hash;
+    let versioned_hashes_hash: Option<[u8; 32]> = input
```
jsign, line 21: Whenever Ethrex supports OpenVM and Zisk, I'll extract Ethrex guest program as I did with Reth to avoid this repetition.

ere-guests/stateless-validator/reth/guest/src/guest.rs
```diff
@@ -31,6 +31,9 @@ pub fn ethereum_guest<S: SDK>() {
     S::cycle_scope(ScopeMarker::Start, "public_inputs_preparation");
     let header = input.block.header().clone();
     let parent_hash = input.block.parent_hash;
+    let versioned_hashes_hash = calculate_versioned_hashes_hash(&input.chain_config, &input.block);
```
jsign, line 34: This is the generic Reth guest program used for all zkVMs. Should feel quite natural what we're doing here.

ere-guests/libs/src/blobs.rs
kevaundray: Noting here that since blob hashes are fixed length, we don't need a length prefix, ie we are not in the situation where we could be hashing ("ab", c") or ("a", "bc") in both cases the hash would be hash("abc") which is ambiguous.

kevaundray: Could we check if most zkVMs have a sha256 precompile? 

(given this grows with number of blobs, it could become quite expensive)

ere-guests/stateless-validator/ethrex/risc0/src/main.rs
```diff
@@ -43,11 +67,17 @@ fn main() {
...
+            env::commit(&parent_beacon_block_root);
+            env::commit(&requests_hash);
             env::commit(&true);
         }
         Err(_) => {
             env::commit(&block_hash.0);
             env::commit(&parent_hash.0);
+            env::commit(&versioned_hashes_hash);
+            env::commit(&parent_beacon_block_root);
+            env::commit(&requests_hash);
```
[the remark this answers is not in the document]
jsign: Yes, and potentially the kzg commitments too if we end up adding the hash commitment as part of EL header.

ere-guests/libs/src/blobs.rs
jsign: > Noting here that since blob hashes are fixed length, we don't need a length prefix, ie we are not in the situation where we could be hashing ("ab", c") or ("a", "bc") in both cases the hash would be hash("abc") which is ambiguous.

Ah yes, forgot to clarify this. This is why I didn't do the "double hashing" done in the EIP-7685 since in that case the `data` isn't fixed length so had that risk.

> Could we check if most zkVMs have a sha256 precompile?

Good call, I'll check and report back.

jsign: Confirmed most support it, for reference it is easier to see [here](https://github.com/eth-act/zkevm-benchmark-workload/blob/master/precompile-patches)

PR #208, opened 2025-10-18
Remove wrapped UncompressedPublicKeys

jsign:
Now that https://github.com/paradigmxyz/reth/pull/19115 was merged we can do this.

### DIFF
--- a/ere-guests/libs/src/senders.rs
+++ b/ere-guests/libs/src/senders.rs
@@ -1,13 +1,7 @@
 //! This module provies sender recovery helpers for witness generation and block stateless validation.
 
 use reth_ethereum_primitives::TransactionSigned;
-use serde::{Deserialize, Serialize};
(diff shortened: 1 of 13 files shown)


ere-guests/libs/src/senders.rs
```diff
@@ -1,13 +1,7 @@
...
 
 use reth_ethereum_primitives::TransactionSigned;
-use serde::{Deserialize, Serialize};
-use serde_with::{Bytes, serde_as};
-
-/// Uncompressed public key wrapper that supports bincode serialization
-#[serde_as]
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct UncompressedPublicKey(#[serde_as(as = "Bytes")] pub [u8; 65]);
+use reth_stateless::UncompressedPublicKey;
```
jsign: cc @kevaundray

PR #210, opened 2025-10-24
Guest crates support `no_std`

han0110:
- Make the following crates `#![no_std]` when `std` feature is off:
  - `guest-libs`
    - In this crate there are `ethereum_ssz` and  `ethereum_ssz_derive` that are `no_std` incompatible, so we need `std` feature to make sure it compiles for customized toolchain that support `std`.
    - For stock rust that need `no_std` we can patch the `ethereum_ssz*` crates latter along with the precompiles, and turn off the `std`
  - `reth-guest` - Used to enable `guest-libs/std`, and to enable `once_cell/std` to avoid missing `critical_section` impl
- Make crate `reth-io` `#![no_std]` because it doesn't have deps that are `no_std` incompatible
- This enables us to add Airbender reth guest (will open a PR after this one and https://github.com/eth-act/ere/pull/178 are merged)

### DIFF
--- a/ere-guests/Cargo.toml
+++ b/ere-guests/Cargo.toml
@@ -179,6 +179,7 @@ serde_json = { version = "*", default-features = false }
 serde_with = { version = "3", default-features = false }
 thiserror = { version = "2.0.0", default-features = false }
 anyhow = { version = "1.0", default-features = false }
+once_cell = { version = "1.21", default-features = false }
 
 ethereum_ssz_derive = "0.9"
 ethereum_ssz = "0.9"
--- a/ere-guests/stateless-validator/reth/guest/Cargo.toml
+++ b/ere-guests/stateless-validator/reth/guest/Cargo.toml
@@ -29,7 +29,8 @@ k256.workspace = true
 bincode_v2.workspace = true
 serde.workspace = true
 serde_with.workspace = true
+once_cell.workspace = true
 
 [features]
 default = ["std", "reth-stateless/default"]
-std = ["guest-libs/std"]
+std = ["guest-libs/std", "once_cell/std"]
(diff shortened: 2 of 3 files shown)


ere-guests/Cargo.toml
```diff
@@ -170,14 +170,16 @@ revm-bytecode = { version = "7.0.2", default-features = false }
 
 # misc
 bincode = "1.3"
+bincode_v2 = { package = "bincode", version = "2.0.1", default-features = false }
```
han0110: To avoid this PR becoming a huge PR, I didn't try to update all bincode to v2, so here we name it like this

ere-guests/stateless-validator/reth/guest/Cargo.toml
```diff
@@ -26,6 +26,11 @@ alloy-primitives.workspace = true
...
 # external
 k256.workspace = true
-bincode.workspace = true
+bincode_v2.workspace = true
 serde.workspace = true
 serde_with.workspace = true
+once_cell.workspace = true
+
+[features]
+default = ["std", "reth-stateless/default"]
```
han0110: Need `reth-stateless/default` enables `k256`.

Also thought about requiring the feature `k256` in each guest crate (which is more straightforward), but that seems easy to forget to add.

crates/benchmark-runner/Cargo.toml
```diff
@@ -8,7 +8,7 @@ witness-generator.workspace = true
 zkevm-metrics.workspace = true
 ere-zkvm-interface.workspace = true
 
-guest-libs.workspace = true
+guest-libs = { workspace = true, default-features = true }
```
han0110: Need this so the `ethereum_ssz` can compile (it's `no_std` incompatible).

PR #211, opened 2025-10-24
Add Airbender stateless validation guest program

han0110:
- Update `ere` to `0.0.14-5c342ee`
- Add crate `reth-airbender-stateless-validator` with patching crates:
  - `radium` - To make it recognize target `riscv32ima` (diff https://github.com/han0110/radium/commit/5644e50792bab60b1ea6dbb0ea7ee555c4b27742)
  - `ethereum_ssz{_derive}` - To make it `no_std` compatible (diff https://github.com/han0110/ethereum_ssz/commit/4a5312b676699bb24ff246a0f522db8c05c6212c)
- Update the xtask script to run `cargo update --package {pkg1} --package {pkg2} ...` before forwarding the cargo command, because I observe that some patching is not taking effect unless we update it explicitly, not sure if there is easier approach to do this.
- To make `reth-airbender-stateless-validator` binary small enough (the maximum is 2MiB required by Airbender)
  - Override `profile.release` with:
    - `lto = true` (allow optimizations across all crates)
    - `panic = "abort"` (not sure whether this would make debugging more difficult or not)
    - `codegen-units = 1` (single unit codegen to get most-performant/smallest binary)
  - Override `profile.release.reth-airbender-stateless-validator` with `opt-level = "s"`
  - The reason to override `profile.release` is because for per package, we couldn't override the `lto` or `panic` or `codegen-units` field, not sure if this is a good for all the other guests or not tho.
    I think a better approach is probably add a new profile, but when trying to compile the guest we'd need to set some env variable for `ere` to know which profile to use

### DIFF
--- a/tests/src/stateless_validator.rs
+++ b/tests/src/stateless_validator.rs
@@ -23,6 +23,7 @@ mod tests {
             (ExecutionClient::Reth, ErezkVM::Risc0),
             (ExecutionClient::Reth, ErezkVM::OpenVM),
             (ExecutionClient::Reth, ErezkVM::Pico),
+            (ExecutionClient::Reth, ErezkVM::Airbender),
             (ExecutionClient::Ethrex, ErezkVM::SP1),
             (ExecutionClient::Ethrex, ErezkVM::Risc0),
             // (ExecutionClient::Ethrex, ErezkVM::OpenVM), // See https://github.com/eth-act/ere/issues/168
@@ -39,6 +40,7 @@ mod tests {
             (ExecutionClient::Reth, ErezkVM::OpenVM),
             (ExecutionClient::Reth, ErezkVM::Pico),
             (ExecutionClient::Reth, ErezkVM::Zisk),
+            (ExecutionClient::Reth, ErezkVM::Airbender),
             (ExecutionClient::Ethrex, ErezkVM::SP1),
             (ExecutionClient::Ethrex, ErezkVM::Risc0),
             // (ExecutionClient::Ethrex, ErezkVM::OpenVM), // See https://github.com/eth-act/ere/issues/168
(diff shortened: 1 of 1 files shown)


ere-guests/Cargo.toml
```diff
@@ -191,6 +192,14 @@ k256 = { version = "0.13", default-features = false, features = [
 ] }
 sparsestate = { git = "https://github.com/eth-act/zkvm-ethereum-mpt.git", tag = "v0.3.0" }
 
+[profile.release]
+lto = true
+panic = "abort"
+codegen-units = 1
+
+[profile.release.package.reth-airbender-stateless-validator]
+opt-level = "s"
```
jsign: Is this a recommendation from airbender?

xtask/src/main.rs
```diff
@@ -74,24 +74,39 @@ fn main() -> Result<()> {
...
+        root_doc["patch"]["crates-io"].as_table_mut().unwrap()[k] = v.clone(); // overwrite / insert
     }
 
     // 4 ── write the updated manifest back
     fs::write(&manifest_path, root_doc.to_string())
         .with_context(|| format!("writing {}", manifest_path.display()))?;
 
-    // 5 ── forward to Cargo
+    // 5 ── run cargo update for patching crates
+    let status = Command::new("cargo")
```
jsign: This was mainly to update Cargo.lock when applying patches instead of when it is first run/compiled?

ere-guests/Cargo.toml
han0110: Ah no, it's because `opt-level = 3` still generate binary > 2MiB, so I have to use either `s` or `z` to optmize for binary size specifically, and `s` is small enough so I picked `s`.

xtask/src/main.rs
han0110: Yes, not sure why the patches of Airbender will be marked as unused in `Cargo.lock` if we don't run `cargo update` explicitly, but other patches seems working fine

PR #221, opened 2025-11-06
CI: aggregate by EL/zkVM

jsign:
This PR aggregates job runs by zkVM/EL to remove the "test" dimension. This is to help the devops infra to reduce the parallelism of jobs by ~75%.

Also, re-enable devops GH mirroring which should help download speed considering now _should_ be more reliable due to controlled parallelism.

Unfortunately, I had to remove the GH page generation with the results since was designed with the previous parallelism strategy. I think this should be fine since we ended up autogenerating other website for runs benchmarks in the other repo.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -74,7 +74,7 @@ jobs:
             threads: 12
           - zkvm: airbender
             el: reth
-            tests: "execute_empty_block execute_mainnet_blocks execute_invalid_block prove_empty_block"
+            tests: "execute_empty_block execute_mainnet_blocks execute_invalid_block"
             threads: 12
 
           # Stateless Validator - Ethrex


.github/workflows/integration-tests.yml
```diff
@@ -45,157 +45,73 @@ jobs:
...
-            zkvm: zisk # See https://github.com/eth-act/ere/issues/186
-          - el: ethrex
-            zkvm: airbender
-          # Pico
+            el: reth
+            tests: "execute_empty_block execute_invalid_block"
+            threads: 12
+          - zkvm: airbender
+            el: reth
+            tests: "execute_empty_block execute_mainnet_blocks execute_invalid_block"
```
jsign: The `execute_mainnet_blocks` has been quite flaky lately, so I decided to remove it.
I want to see if we convinced now devops to give us maybe a bit more powerful machine if this can improve, or maybe the flakiness is related to Airbender SDK..

PR #231, opened 2025-12-08
Update Ere

han0110:
- Update `ere` to `2ede04d` with the new `Input` structure
- Refactor the `GuestFixture` a bit
- Update guest regarding to the `Platform` update, and use the `OutputHashedPlatform`
- Update OpenVM to `v1.4.2`, but sadly the [`openvm-reth-benchmark`](https://github.com/axiom-crypto/openvm-reth-benchmark) hasn't had the new version so we have to patch the OpenVM to the same version.


jsign (approved):
LGTM

PR #259, opened 2026-02-10
Support Zisk profiling

jsign:
This PR adds a feature to dump Zisk profiles when running in `execute` mode.

If the `--zisk-profile` flag is added when using `--action execute`, it will dump `ziskemu` profiles of all executed fixtures in an output folder (`zisk-profiles` by default or can be overridden by `--zisk-profile-output`).

```
> tree zisk-profiles               
zisk-profiles
└── ethrex-v9.0.0
    ├── zisk_profile_rpc_block_23533518.prof
    ├── zisk_profile_rpc_block_23533705.prof
    ├── zisk_profile_rpc_block_23533763.prof
    ├── zisk_profile_rpc_block_23533796.prof
...
```

I also added a Python script that can read this output folder and provide some summary about all the fixture runs regarding costs (e.g. avg, min, max, etc.), and also a plot to have an easier visual on this. I'll soon share a HackMD doc with results from 500 mainnet blocks.

### DIFF
--- a/crates/benchmark-runner/src/runner.rs
+++ b/crates/benchmark-runner/src/runner.rs
@@ -71,19 +71,16 @@ pub fn run_benchmark(
 ) -> Result<()> {
     HardwareInfo::detect().to_path(config.output_folder.join("hardware.json"))?;
 
-    let zisk_elf = config
-        .zisk_profile_config
-        .is_some()
-        .then_some(instance.elf.as_slice());
     let zkvm = &instance.zkvm;
+    let elf = &instance.elf;
     match config.action {
         Action::Execute => inputs
             .into_par_iter()
-            .try_for_each(|input| process_input(zkvm, input, config, zisk_elf))?,
+            .try_for_each(|input| process_input(zkvm, input, config, elf))?,
 
         Action::Prove => inputs
             .into_iter()
-            .try_for_each(|input| process_input(zkvm, input, config, zisk_elf))?,
+            .try_for_each(|input| process_input(zkvm, input, config, elf))?,
     }
 
     Ok(())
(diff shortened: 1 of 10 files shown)


crates/benchmark-runner/src/runner.rs
```diff
@@ -13,10 +13,30 @@ use ere_zkvm_interface::{zkVM, ProofKind, ProverResourceType};
...
 use crate::guest_programs::{GuestFixture, OutputVerifierResult};
+use crate::zisk_profiling::run_profiling;
+
+pub use crate::zisk_profiling::ProfileConfig;
 
 /// Default version tag for guest programs
 const DEFAULT_GUEST_VERSION: &str = "v0.5.0";
 
+/// A zkVM instance bundled with ELF bytes (used for profiling).
+pub struct ZkVMInstance {
```
jsign: Before we moved just `DockerizedzkVM` between the setup and run phase. Now we move `ZkVMInstance` which also contains the ELF needed if the new flag is activated.

crates/benchmark-runner/src/stateless_validator.rs
```diff
@@ -92,7 +93,7 @@ pub fn reth_inputs_from_fixture(
     fixtures: &[StatelessValidationFixture],
 ) -> Result<Vec<Box<dyn GuestFixture>>> {
     fixtures
-        .iter()
+        .par_iter()
```
jsign: Unrelated improvement: parallelize input generation from fixtures. Helps to save some seconds if we have many fixtures to run.

scripts/analyze_zisk_profiles.py
jsign: Not super interesting to review.

han0110 (approved):
LGTM!

scripts/analyze_zisk_profiles.py
dionatasvieira498-source: Preciso que me dê a liberação

PR #270, opened 2026-03-31
Update ere and add support for timeouts

jsign:
Updates ere, particularly to include the [new timeout feature](https://github.com/eth-act/ere/pull/324).


crates/ere-hosts/src/main.rs
```diff
@@ -12,17 +12,22 @@ use benchmark_runner::{
...
 use tracing::info;
 use tracing_subscriber::EnvFilter;
 
 use crate::cli::{Cli, GuestProgramCommand};
 
 pub mod cli;
 
+const DEFAULT_EXECUTE_TIMEOUT: Duration = Duration::from_secs(5 * 60);
+const DEFAULT_PROVE_TIMEOUT: Duration = Duration::from_secs(15 * 60);
+const DEFAULT_VERIFY_TIMEOUT: Duration = Duration::from_secs(2);
```
jsign, line 29: I think these are reasonable defaults, but the user can override with `--timeout` that will be interpreted depending on which `--action` is performed.

PR #279, opened 2026-04-23
Remove using Reth stateless fork & other misc improvements

jsign:
## Summary

Updates the witness generation stack and improves EEST fixture generation so large fixture sets can be generated more reliably.

## Changes

- Bumps Reth dependencies to `v2.1.0`, Alloy dependencies to `2.0`, and updates `ef-tests` / `stateless` to the newer upstream revision.
- Removes the local `jsign/stateless` patch override.
- Refactors EEST fixture loading to filter raw test names before deserializing selected cases.
- Adds an EEST-specific `generate_to_path` implementation that streams fixtures to disk instead of retaining every generated fixture in memory.
- Runs matching EEST cases in parallel and skips cases that fail before producing an executable block witness.
- Adds test case names to EEST execution errors for easier debugging.
- Updates raw-input and RPC fixture generation for the dependency API changes.
- Documents EEST parallelism, skipped-case behaviour, and the updated fixture generation command.
- Remove `block-encoding-length` fixture generation
- Pull ELFs from Tag or Commit from `ere-guests` instead of maintaining a local hardcoded target.


README.md
```diff
@@ -37,7 +37,7 @@ cargo run -p ere-hosts -- --help
 Generate sample fixtures into `zkevm-fixtures-input/`:
 
 ```bash
-cargo run -p witness-generator-cli --release -- tests --include 10M --include Prague
+EF_TEST_TRIE=default RUST_MIN_STACK=16388608 RUST_LOG=info RAYON_NUM_THREADS=8 cargo run -p witness-generator-cli --release -- tests
```
jsign, line 40: Clarifying a bit some env vars that `paradigmxyz/stateless` still needs. Not nice but at least I can remove now using a fork.

With Kev we are trying to see if we can fill execution witness during EEST benchmark generation -- when that day comes, we can remove basically the whole `witness-generator-cli` stuff for EEST that requires `stateless` for "out of band" execution witness generation.

PR #281, opened 2026-04-27
feat: add zisk cluster benchmark support

han0110:
- Bump `ere` to `44e91ed7c0bd5398d55eb4ce43594ac0f161dd2b`
- Add dep `ere-cluster-client-zisk` so we can use `ZiskClusterClient` to do the proving benchmark
  Although `DockerizedzkVM` also supports `ProverResource::Cluster`, but it requires additional configuration for the container to get access to the actual cluster (extra hosts), so this PR integrates the client directly to avoid that.
- Make the `ZkVMInstance` an enum, and `Dockerized` variant works as is. For `ZiskClusterClient` it supports `prove` and `verify` (by integrating the verifier directly in the cluster client crate), but it doesn't support `execute`

### DIFF
--- a/crates/benchmark-runner/src/runner.rs
+++ b/crates/benchmark-runner/src/runner.rs
@@ -1,10 +1,14 @@
 //! Runner for benchmark tests
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
+use ere_cluster_client_zisk::{ZiskClusterClient, ZiskProgramVk, ZiskProof};
 use ere_dockerized::{
-    zkVMKind, DockerizedzkVM, DockerizedzkVMConfig, Elf, EncodedProof, ProverResource,
+    codec::{Decode, Encode},
+    zkVMKind, zkVMVerifier, DockerizedzkVM, DockerizedzkVMConfig, Elf, EncodedProof, Input,
+    ProgramExecutionReport, ProgramProvingReport, ProverResource, PublicValues,
 };
-use ere_guests_downloader::Downloader;
+use ere_guests_downloader::{CompiledGuest, Downloader};
+use ere_util_tokio::block_on;
 use rayon::iter::{ParallelBridge, ParallelIterator};
 use std::fs;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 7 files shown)


crates/benchmark-runner/src/runner.rs
```diff
@@ -300,27 +381,51 @@ pub async fn get_guest_zkvm_instances(
...
 
-async fn load_elf(guest_name: &str, bin_path: Option<&Path>) -> Result<Vec<u8>> {
+async fn load_compiled(guest_name: &str, bin_path: Option<&Path>) -> Result<CompiledGuest> {
     if let Some(path) = bin_path {
-        return fs::read(path.join(format!("{guest_name}.elf")))
-            .with_context(|| format!("Failed to read ELF from path: {}", path.display()));
+        let elf = fs::read(path.join(format!("{guest_name}.elf")))
+            .with_context(|| format!("Failed to read ELF from path: {}", path.display()))?;
+        let program_vk = fs::read(path.join(format!("{guest_name}.vk")))
+            .with_context(|| format!("Failed to read program vk from path: {}", path.display()))?;
```
han0110, line 420: This might be a breaking change if the `bin_path` didn't contain the program vk.

crates/benchmark-runner/src/runner.rs
```diff
@@ -25,23 +29,107 @@ const ERE_GUESTS_DOWNLOAD_KIND: &str = env!("ERE_GUESTS_DOWNLOAD_KIND");
...
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("ZkVMInstance")
-            .field("zkvm", &self.zkvm.name())
-            .field("elf_len", &self.elf().len())
-            .finish()
+        match self {
+            Self::Dockerized(vm) => f
+                .debug_struct("Dockerized")
+                .field("zkvm", &vm.name())
+                .field("elf", vm.elf())
```
han0110, line 125: The `Elf` implements `Debug` to only prints the length now, so we can skip doing it manually

crates/benchmark-runner/src/runner.rs
jsign: I think it should be fine.

jsign (approved):
LGTM!

PR #284, opened 2026-05-13
Update to ere@v0.9.1 and ere-guests@v0.10.0

jsign:


crates/benchmark-runner/src/runner.rs
```diff
@@ -192,7 +192,7 @@ where
 }
 
 fn benchmark_zkvm_name(zkvm: &ZkVMInstance) -> String {
-    format!("{}-v{}", zkvm.name(), zkvm.sdk_version())
```
jsign, line 195: https://github.com/eth-act/ere/pull/353#issuecomment-4385611805

PR #285, opened 2026-05-21
feat: update to ere@v0.11.0 and ere-guests@v0.11.0

han0110:
- ZisK gets updated to v0.18.0, and the profiling output of `ziskemu` change a bit, an sample `.prof` file can be found [here](https://gist.github.com/han0110/9b718ef5a44485a8d6ae725e836b4097#file-zisk_profile_rpc_block_24950000-prof), and a sample report generated can be found [here](https://gist.github.com/han0110/9b718ef5a44485a8d6ae725e836b4097#file-report-md). Note that the `Cost Breakdown by Scope` section now doesn't show `Opcodes | Precompiles | Memory` columns because `ziskemu` drop the support of that of profile tags (user defined cycle scope)
- `Platform` API is updated to be compatible with `zkvm-standard` so no longer need to do `with_prefixed_stdin`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@ zkevm-fixtures*
 zkevm-metrics*
 # zkevm dump
 zkevm-dump
+# zisk profiles
+zisk-profiles
 
 # vscode settings
 .vscode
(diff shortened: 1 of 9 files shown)


jsign (approved):
Great 🙏 
LGTM!

PR #286, opened 2026-05-21
feat: add support for EEST blockchain test fixtures

jsign:
This PR:
- Adds support in `ere-hosts` to directly consume EEST tests new format from `execution-specs@projects/zkevm`. It is the usual fixture format, but for each block contains `statelessInputBytes` and `statelessOutputBytes`.
- The runner automatically can detect both workload fixture formats: current one (now called `Legacy`, and new one `Eest`). i.e. this is not a breaking change.
- If `Eest` fixtures are detected, we directly use `stateless{Input|Output}Bytes` for the guest program run (i.e. running EEST fixtures can be done directly without a previously required `witness-generator-cli`).
- The `zkevm-metrics` metadata for an EEST run contains more details to easily map to original EEST file.

Example run output in `zkevm-metrics`:
```
{
  "name": "eest__tests_amsterdam_eip8025_optional_proofs_test_witness_state_writes_py_test_witness_state_reverted_sstore_still_contains_storage_proof_fork_Amsterdam-blockchain_test__block0",
  "timestamp_completed": "2026-05-21T17:48:02.645237188Z",
  "metadata": {
    "block_index": 0,
    "block_number": 1,
    "block_used_gas": 26012,
    "chain_id": 1,
    "fixture_format": "eest",
    "network": "Amsterdam",
    "original_test_name": "tests/amsterdam/eip8025_optional_proofs/test_witness_state_writes.py::test_witness_state_reverted_sstore_still_contains_storage_proof[fork_Amsterdam-blockchain_test]",
    "source_path": "blockchain_tests/for_amsterdam/amsterdam/eip8025_optional_proofs/witness_state_writes/witness_state_reverted_sstore_still_contains_storage_proof.json"
  },
  "execution": {
    "crashed": {
      "reason": "zkVM method error: Emulator panicked: Mem::read() section not found for addr: 0=0 with width: 8"
    }
  }
}
```

### DIFF
--- a/crates/benchmark-runner/src/stateless_validator.rs
+++ b/crates/benchmark-runner/src/stateless_validator.rs
@@ -1,25 +1,16 @@
 //! Stateless validator guest program.
 
-use crate::guest_programs::{GenericGuestFixture, GuestFixture};
-use anyhow::{bail, Context, Result};
-use ere_guests_guest::Guest;
-use ere_guests_integration_tests::NoopPlatform;
-use ere_guests_stateless_validator_ethrex::{
-    guest::StatelessValidatorEthrexGuest,
-    host::{build_eip8025_input, Eip8025InputSource},
-};
-use ere_guests_stateless_validator_reth::guest::{
-    StatelessValidatorRethGuest, StatelessValidatorRethInput,
-};
+mod eest;
+mod fixtures;
+mod inputs;
+
+use crate::guest_programs::GuestFixture;
+use anyhow::Result;
 use serde::{Deserialize, Serialize};
-use std::{
-    collections::HashSet,
-    path::{Path, PathBuf},
-};
+use std::path::Path;
 use strum::{AsRefStr, EnumString};
-use tracing::info;
-use walkdir::WalkDir;
-use witness_generator::StatelessValidationFixture;
+
+pub use fixtures::{benchmark_fixture_paths, iter_benchmark_fixture_paths, load_benchmark_fixture};
 
 /// Execution client variants.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, EnumString, AsRefStr)]
@@ -49,48 +40,6 @@ impl ExecutionClient {
     }
 }
 
-/// Lazily walks a fixture folder and yields each fixture file path.
-pub fn iter_benchmark_fixture_paths(path: &Path) -> impl Iterator<Item = PathBuf> {
-    WalkDir::new(path)
-        .min_depth(1)
-        .into_iter()
-        .filter_map(|entry| entry.ok())
-        .filter(|entry| entry.file_type().is_file())
-        .map(walkdir::DirEntry::into_path)
-}
-
-/// Resolves benchmark fixture file paths, optionally filtering by fixture name prefix.
-pub fn benchmark_fixture_paths(
-    input_folder: &Path,
-    selected_fixtures: Option<&[String]>,
-) -> Result<Vec<PathBuf>> {
-    let available_paths: Vec<_> = iter_benchmark_fixture_paths(input_folder).collect();
-    let Some(selected_fixtures) = selected_fixtures.filter(|fixtures| !fixtures.is_empty()) else {
-        return Ok(available_paths);
-    };
-
-    let fixture_prefixes = normalize_fixture_prefixes(selected_fixtures)?;
-    let mut resolved_paths = Vec::new();
-
-    for path in available_paths {
-        let fixture_name = fixture_name_from_path(&path)?;
-        if fixture_prefixes
-            .iter()
-            .any(|prefix| fixture_name.starts_with(prefix))
-        {
-            resolved_paths.push(path);
-        }
-    }
-
-    Ok(resolved_paths)
-}
-
-/// Reads and deserializes a single benchmark fixture file.
-pub fn load_benchmark_fixture(path: &Path) -> Result<StatelessValidationFixture> {
-    let content = std::fs::read(path)?;
-    serde_json::from_slice(&content).with_context(|| format!("Failed to parse {}", path.display()))
-}
-
 /// Lazily prepares stateless validator inputs from a fixture folder.
 pub fn stateless_validator_input_iter(
     input_folder: &Path,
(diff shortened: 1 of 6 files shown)


crates/benchmark-runner/src/stateless_validator/eest.rs
jsign, line 1: The previous `crates/benchmark-runner/src/stateless_validator.rs` file was getting a bit loaded. I moved to a module with a bit more organized structure.

I'll simply comment on the new stuff.

crates/benchmark-runner/src/stateless_validator/eest.rs
```diff
@@ -0,0 +1,352 @@
+use anyhow::{bail, Context, Result};
+use serde::Deserialize;
+use sha2::{Digest, Sha256};
+use std::{collections::BTreeMap, path::Path};
+
+const EEST_SAFE_NAME_MAX_LEN: usize = 80;
+
+#[derive(Debug, Clone)]
+pub(crate) struct EestStatelessFixture {
```
jsign, line 12: Now we have an enum for `BenchmarkFixture`:
```
#[derive(Debug, Clone)]
pub(crate) enum BenchmarkFixture {
    Legacy(StatelessValidationFixture),
    Eest(EestStatelessFixture),
}
```

`StatelessValidationFixture` is the usual format. This new `EestStatelssFixture` is the one coming from EEST variant. It mainly contains basic data extracted from parsing EEST fixture that will be also used as `Metadata` for `zkevm-metrics` outputs.

Example of output:
```
{
  "name": "eest__tests_amsterdam_eip8025_optional_proofs_test_witness_state_writes_py_test_witness_state_reverted_sstore_still_contains_storage_proof_fork_Amsterdam-blockchain_test__block0",
  "timestamp_completed": "2026-05-21T17:48:02.645237188Z",
  "metadata": {
    "block_index": 0,
    "block_number": 1,
    "block_used_gas": 26012,
    "chain_id": 1,
    "fixture_format": "eest",
    "network": "Amsterdam",
    "original_test_name": "tests/amsterdam/eip8025_optional_proofs/test_witness_state_writes.py::test_witness_state_reverted_sstore_still_contains_storage_proof[fork_Amsterdam-blockchain_test]",
    "source_path": "blockchain_tests/for_amsterdam/amsterdam/eip8025_optional_proofs/witness_state_writes/witness_state_reverted_sstore_still_contains_storage_proof.json"
  },
  "execution": {
    "crashed": {
      "reason": "zkVM method error: Emulator panicked: Mem::read() section not found for addr: 0=0 with width: 8"
    }
  }
}
```

crates/benchmark-runner/src/stateless_validator/eest.rs
```diff
@@ -0,0 +1,352 @@
...
+    pub(crate) block_index: usize,
+    pub(crate) network: String,
+    pub(crate) chain_id: u64,
+    pub(crate) block_number: Option<u64>,
+    pub(crate) block_used_gas: Option<u64>,
+    pub(crate) stateless_input_bytes: Vec<u8>,
+    pub(crate) stateless_output_bytes: Vec<u8>,
+}
+
+#[derive(Debug, Deserialize)]
```
jsign, line 25: EEST fixture parsing structs.

crates/benchmark-runner/src/stateless_validator/fixtures.rs
```diff
@@ -0,0 +1,340 @@
...
+use walkdir::WalkDir;
+use witness_generator::StatelessValidationFixture;
+
+const EEST_BLOCKCHAIN_TESTS_DIR: &str = "blockchain_tests";
+
+#[derive(Debug, Clone)]
+pub(crate) enum BenchmarkFixture {
+    Legacy(Box<StatelessValidationFixture>),
+    Eest(EestStatelessFixture),
+}
```
jsign, line 27: Note now `BenchmarkFixture` is an enum.

crates/benchmark-runner/src/stateless_validator/fixtures.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    if value.get("stateless_input").is_some() {
+        let fixture = serde_json::from_value(value)
+            .with_context(|| format!("Failed to parse legacy fixture {}", path.display()))?;
+        return Ok(vec![BenchmarkFixture::Legacy(Box::new(fixture))]);
+    }
+
+    load_eest_benchmark_fixtures(value, path, input_root)
+        .map(|fixtures| fixtures.into_iter().map(BenchmarkFixture::Eest).collect())
+}
```
jsign, line 198: The automatic detection if we are in the legacy or eest format.

crates/benchmark-runner/src/stateless_validator.rs
jsign, line 1: As said, mostly refactoring to make it more modular.

crates/benchmark-runner/src/stateless_validator/inputs.rs
```diff
@@ -0,0 +1,120 @@
...
+    el: ExecutionClient,
+) -> Result<Box<dyn GuestFixture>> {
+    match fixture {
+        BenchmarkFixture::Legacy(fixture) => match el {
+            ExecutionClient::Reth => reth_input_from_fixture(*fixture),
+            ExecutionClient::Ethrex => ethrex_input_from_fixture(*fixture),
+        },
+        BenchmarkFixture::Eest(fixture) => raw_eest_input_from_fixture(fixture),
+    }
+}
```
jsign, line 46: Gist-ish of the PR: depending on the `BenchmarkFixture` variant, we do the usual for `Legacy`, or call new `raw_eest-input_from_fixture(...)`.

crates/benchmark-runner/src/stateless_validator/inputs.rs
```diff
@@ -0,0 +1,120 @@
...
+        fixture_format: "eest",
+        original_test_name: fixture.original_test_name,
+        source_path: fixture.source_path,
+        block_index: fixture.block_index,
+        network: fixture.network,
+        chain_id: fixture.chain_id,
+        block_number: fixture.block_number,
+        block_used_gas: fixture.block_used_gas,
+    };
+    let expected_public_values = Sha256::digest(fixture.stateless_output_bytes).to_vec();
```
jsign, line 59: Nit: here we explicitly `Sha256` for now. But I think in theory, zkVMs are expected to output raw bytes that the guest program wrote to output.

@han0110, do you know if this is something we could assume in new Zisk v0.18.0? Or maybe we should still keep this Sha256-ing for a while.

crates/benchmark-runner/src/stateless_validator/fixtures.rs
```diff
@@ -0,0 +1,340 @@
...
+                .components()
+                .any(|component| component.as_os_str() == ".meta")
+        })
+        .map(walkdir::DirEntry::into_path)
+}
+
+/// Resolves benchmark fixture file paths.
+pub fn benchmark_fixture_paths(
+    input_folder: &Path,
+    _selected_fixtures: Option<&[String]>,
```
han0110: seems to be leftover?

crates/benchmark-runner/src/stateless_validator/inputs.rs
han0110: > But I think in theory, zkVMs are expected to output raw bytes that the guest program wrote to output.

Yes but currently only SP1 and RISC Zero support that, others have restricted size:

- Airbender - 32 bytes
- OpenVM - 32 bytes (configurable)
- ZisK - 256 bytes

So I think we have to stick with the sha256 output for a while, or we drop the support of Airbender in `ere-guests` and configure OpenVM to support 256 bytes as well (assuming the public values will not exceed 256 bytes, where currently it's 41 bytes)

han0110 (approved):
LGTM!  (looks like the fixture name is a bit tricky to handle when needing to use it for file name)

crates/benchmark-runner/src/stateless_validator/fixtures.rs
jsign: Yep, thanks for the catch.

crates/benchmark-runner/src/stateless_validator/inputs.rs
jsign: Cool, let's keep the `Sha256`, and we can see how things evolve.

@marcinbugaj, is this something Zisk is planning to support? Since I think from zkEVM standards POV it should be required the return exactly what the guest program wrote as output?

jsign:
>  (looks like the fixture name is a bit tricky to handle when needing to use it for file name)

Yeah, it is a bit tricky not to pass 256 in length. Before, I used a shorter style but had to add a hash to the name to avoid potential collisions, which was ugly. _I think_ it should be fine. If we discover it violates the length max, we should prob switch to that idea too.
For UI-related stuff, it should use the `metadata` field in the JSON result, which is the original "nice" name (i.e., not rely on the fixture result file name).

jsign:
Might leave this open for some extra mins/hours -- I'm using this branch to test a dashboard automation reg EEST fixtures direct run.

crates/benchmark-runner/src/stateless_validator/inputs.rs
marcinbugaj: I think that ZisK already supports access to original data committed to by `write_output`. In their `sha_hasher` example, in `prove.rs` one has:
```
proof.with_program_vk(&vk).with_publics(&publics).verify()?;
```
where `publics` has type:
```
#[derive(Default, Debug, Serialize, Deserialize)]
pub struct PublicValues {
    data: Vec<u8>,
    #[serde(skip)]
    ptr: AtomicUsize,
}
```
It seems that `PublicValues` contains unaltered data passed to `write_output`. That's for 0.18.0 release

Also, have a look at `execute.rs`:
```
let output: Output = result.get_public_values_abi()?;
```
where Output is a data type commited to in `guest/src/main.rs`

Source: https://github.com/0xPolygonHermez/zisk_template

han0110: Right ZisK outputs the bytes as how it's written in guest program. 

The main concern is there is a maximum size of the public values, currently ZisK only support 256 bytes at most (https://github.com/0xPolygonHermez/zisk/blob/main/ziskos/entrypoint/src/lib.rs#L281), although it's pretty enough for our usecase for now, but if we need to bump it to more than 256 bytes some day, we might have an issue and need to go back to hash the output before write it.

(Airbender and OpenVM have the same issue and they only support up to 32 bytes, we are mainly blocked by them to not just output our 41 raw bytes, but as mentioned above, for OpenVM we can configure it to be bigger, for Airbender we could probably drop support of it)

marcinbugaj: You are right. I created an issue in ZisK: https://github.com/0xPolygonHermez/zisk/issues/1033 - they don't conform to the spec

PR #290, opened 2026-05-28
Add experimental witness-generator-spec-cli

jsign:
This PR adds a new experimental `witness-generator-spec-cli` which builds guest program input for live networks using the spec format.

Notes:
- I created it separately from `witness-generator{-cli}` stuff to not mix existing stuff. I think at some point they can be merged, or probably replace the old one, since ideally we should stop using the old format. 
- We leverage as many structs from `stateless-validator-common` as possible. Still, some were missing and have now been added in `schema.rs`. Not sure where this will live eventually -- we could move the `schema.rs` stuff to `ere-guests` eventually so both `workload` and `zkboost` can leverage them as a neutral definition of things.
- It now uses both a CL and an EL RPC, compared to the previous EL-only RPC. 

Regarding the last bullet, I think my goal is to have super-simple code for stateless input generation. Using EL-only RPC means having to do some gymnastics that we already suffered in `ere-guests` reg having to re-execute statelessly to get some fields (e.g. `requests`).

I think I would prefer to avoid all those complex things for two reasons:
- For now, only stateless Reth provides those generated results. I think we should avoid as much as possible, depending on ELs in particular, for our tooling. EL and CL RPC are neutral; we can switch to different implementations, and they have a spec. 
- Avoid re-executing the block to generate inputs.

The current implementation is super simple, mainly does JSON-RPC calls, which usually 1:1 map into our neutral structs, the only exceptions are:
- `StatelessInputs.public_keys`, since it is a field invented for stateless guests and not part of RPCs
- The `versioned_hashes` aren't part of CL RPC. From what I've seen, kzg blobs commitments aren't part of CL RPC in Gloas, which looks a bit surprising to me. I would like to dig a bit more into whether there was a reason for this, since in Osaka RPC I think they were returned.

For both cases, we deserialise the RLP transactions and derive both things, which is quite simple anyway.

Next steps are I'm waiting for `glamsterdam-devnet-4` to be healthy, and would like to test this out in that devnet. If this works correctly, we could incorporate this experiment more into `workload` to generate live network spec guest inputs for live networks (prob only Glamsterdam devnets).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2781,6 +2781,11 @@ dependencies = [
 
 [[package]]
 name = "ere-codec"
+version = "0.11.0"
+source = "git+https://github.com/eth-act/ere?tag=v0.11.0#e1d345c5e4e2b41d1334421310443306ea44f65f"
+
+[[package]]
+name = "ere-codec"
 version = "0.12.1"
 source = "git+https://github.com/eth-act/ere?tag=v0.12.1#628fa3c49941704d72e7b0e038e7bce5b6d2cd58"
 
@@ -2825,6 +2830,11 @@ dependencies = [
 
 [[package]]
 name = "ere-platform-core"
+version = "0.11.0"
+source = "git+https://github.com/eth-act/ere?tag=v0.11.0#e1d345c5e4e2b41d1334421310443306ea44f65f"
+
+[[package]]
+name = "ere-platform-core"
 version = "0.12.1"
 source = "git+https://github.com/eth-act/ere?tag=v0.12.1#628fa3c49941704d72e7b0e038e7bce5b6d2cd58"
 
@@ -2837,7 +2847,7 @@ dependencies = [
  "auto_impl",
  "bincode 2.0.1",
  "clap",
- "ere-codec",
+ "ere-codec 0.12.1",
  "ere-verifier-core",
  "indexmap 2.14.0",
  "serde",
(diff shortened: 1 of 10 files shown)


jsign:
Tested with `glamsterdam-devnet-5` and it seems to be working well.

The network is quite unstable tbh, but had success cases so will be merging soon. Prob will keep iterating on this new crate.

jsign:
```
> cargo run -p witness-generator-spec-cli --release -- \
  --cl-url "$CL_URL" \
  --el-url "$EL_URL" \
  --block-id head
    Finished `release` profile [optimized] target(s) in 0.39s
     Running `target/release/witness-generator-spec-cli --cl-url 'http://127.0.0.1:3500' --el-url 'http://127.0.0.1:8545' --block-id head`
0x000110000000e1830000db3d07001f3e07002c0000008583000033f67f40ee2091e847827f6a8f986b5cfd591f48cdf842acbca155fc9ef6a276c5830000eecb18af972b2e230aad5dd62bf483a73e2535fba7bf1519b976861de883c7edf97e180c050e5ab072211ad2c213eb5aee4df134185308dc7765a8452ec70b12a8e2a9e76d96b6dd72a19db725e929651bfc493650ea9c3e1e1dfa37574a18aa716d75f033c022bcb9886a07b973115ff910727d0020c0904420010000420000804620900080020102006420014249001d0ea020468804008083200001200a813412800940400400000020050003024002044228c682d0850c0481080024278842002024c9001000028200000000204c815000001000400002000801240000a080040810d54003051d3500100020001000885020000c040002880042110000101004420018044805b000000800148040000444c0004400110132000002c04000f000040a20021400060800100400026800002000810000020045004001890801000040a200040010000100360004800000823440304628020000004902000001001030200908010200440140000404000200002028aff196ef8f7892ebbd311d1f21a8320b4a61af839d4df774632ffc562f0e026e7b0000000000 ...
```

PR #292, opened 2026-06-09
feat: update ere and ere-guests

han0110:
- Bump `ere` to `v0.12.1`
- Bump `ere-guests` to `v0.12.1`
- Add flag `--max-concurrency` to `witness-generator-cli rpc` subcommand to download live block concurrently to not fall behind the chain (default to 1 as the original behavior)

### DIFF
--- a/crates/witness-generator-cli/src/main.rs
+++ b/crates/witness-generator-cli/src/main.rs
@@ -82,7 +82,7 @@ enum SourceCommand {
         #[arg(long, value_name = "PATH")]
         genesis: Option<PathBuf>,
 
-        /// Maximum number of blocks and witnesses fetched concurrently
+        /// Maximum number of blocks fetched concurrently while catching up, only applies to `--follow`
         #[arg(long, default_value_t = 1)]
         max_concurrency: usize,
     },
--- a/crates/witness-generator-cli/src/main.rs
+++ b/crates/witness-generator-cli/src/main.rs
@@ -81,6 +81,10 @@ enum SourceCommand {
         /// Optional path to a geth-style genesis.json file for custom/devnet chain config
         #[arg(long, value_name = "PATH")]
         genesis: Option<PathBuf>,
+
+        /// Maximum number of blocks and witnesses fetched concurrently
+        #[arg(long, default_value_t = 1)]
+        max_concurrency: usize,
     },
 }
 
@@ -153,8 +157,10 @@ async fn build_generator(source: SourceCommand) -> Result<Box<dyn FixtureGenerat
             rpc_header,
             genesis,
             follow: listen,
+            max_concurrency,
         } => {
-            let mut builder = RpcBlocksAndWitnessesBuilder::new(rpc_url);
+            let mut builder =
+                RpcBlocksAndWitnessesBuilder::new(rpc_url).max_concurrency(max_concurrency);
 
             if let Some(rpc_header) = rpc_header {
                 let headers = RpcFlatHeaderKeyValues::new(rpc_header)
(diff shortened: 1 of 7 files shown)


crates/witness-generator/src/rpc_generator.rs
```diff
@@ -209,6 +223,37 @@ impl FixtureGenerator for RpcFixtureGenerator {
...
+    /// * `number` - The block number or tag to fetch
+    /// * `full` - Whether to include full transaction bodies
+    ///
+    /// # Errors
+    /// Returns an error if the RPC call fails or if the block cannot be found.
+    async fn fetch_block_by_number(
+        &self,
+        number: BlockNumberOrTag,
+        full: bool,
+    ) -> Result<Block<TransactionSigned>> {
```
han0110, line 238: Tried to DRY the code a bit, here it's hardcoded with `TransactionSigned` and for block header only request (`full = false`) it doesn't affect the decoding.

crates/witness-generator/src/rpc_generator.rs
```diff
@@ -221,75 +266,62 @@ impl RpcFixtureGenerator {
...
+                .number
+                .saturating_sub(last_n_blocks as u64 - 1),
             latest_block.header.number,
         );
 
-        let mut hashes = Vec::with_capacity(last_n_blocks);
-        hashes.push((latest_block.header.number, latest_block.header.hash));
-        for n in (block_num_start..block_num_end).rev() {
-            let block_hash = hashes.last().unwrap().1;
+        let mut blocks: Vec<Block<TransactionSigned>> = Vec::with_capacity(last_n_blocks);
```
han0110: Here it avoids to fetch same block twice.

crates/witness-generator-cli/src/main.rs
```diff
@@ -81,6 +81,10 @@ enum SourceCommand {
         /// Optional path to a geth-style genesis.json file for custom/devnet chain config
         #[arg(long, value_name = "PATH")]
         genesis: Option<PathBuf>,
+
+        /// Maximum number of blocks fetched concurrently while catching up, only applies to `--follow`
+        #[arg(long, default_value_t = 1)]
+        max_concurrency: usize,
```
han0110, line 87: When block and witness fetching time is longer than slot time, the delay to latest block accumulates, eventually the requested block too deep causes the execution witness timed out and stuck there forever.

This allow `--follow` to catch up the chain when falling behind a bit, and it only applies when `--follow` is specified.

crates/witness-generator/src/rpc_generator.rs
```diff
@@ -340,31 +361,42 @@ impl RpcFixtureGenerator {
...
-        for n in block_num..=latest_block.header.number {
-            bws.push(self.fetch_specific_block(n).await?);
+        while let Some((n, result)) = stream.next().await {
+            match result {
+                Ok(bw) => bws.push(bw),
+                Err(err) => {
+                    error!("Fetch block {n} failed: {err}");
+                    break;
+                }
+            }
```
han0110, line 399: Here we return the continous blocks and witnesses we've fetched since `block_num` even the later request failed.

crates/witness-generator/src/rpc_generator.rs
```diff
@@ -433,12 +461,15 @@ impl RpcFixtureGenerator {
...
-                _ = tokio::time::sleep(std::time::Duration::from_secs(6)) => {
+            // Throttle the remaining time up to the minimum interval.
+            let remaining = MIN_LOOP_INTERVAL.saturating_sub(iter_start.elapsed());
+            if !remaining.is_zero() {
+                tokio::select! {
+                    _ = stop_signal.cancelled() => {
+                        info!("Stopped listening for new blocks.");
+                        break;
+                    }
+                    _ = tokio::time::sleep(remaining) => {}
```
han0110, line 472: Previously we always sleep for 6 seconds, which contributes to the delay to the chain head even we are already falling behind a bit. This PR updates the 6s to be the min time of an iteration.

jsign (approved):

PR #296, opened 2026-07-12
Update to ere-guests v0.13

jsign:
Note: I think soon ere-guest might change to `zkevm-tests@v0.6.1` and we'll update `ere-guests` again with a bit more simplifications, but this I think contains bulk of pending changes.

## Summary

  Update the benchmark stack to Ere and Ere Guests v0.13 and align stateless-validator inputs, outputs, and guest
  availability with the new release.

  - Bump Ere and Ere Guests dependencies to v0.13 and update the lockfile.
  - Use the canonical stateless input types and SSZ encoding from `stateless-validator-common`, removing the
  duplicated local schema.
  - Pass canonical EEST inputs and raw SSZ outputs directly to the Reth and Ethrex guests.
  - Restrict legacy `stateless_input` fixtures to Zilkworm.
  - Validate that Zesu receives Glamsterdam inputs and use its published ZisK artifact by default.
  - Report the EL versions associated with the published guest artifacts.
  - Remove the unsupported `empty-program` benchmark command and implementation.
  - Update benchmark and fixture-generation documentation for the new compatibility matrix.

  ## Compatibility changes

  - Reth and Ethrex now require canonical EEST `blockchain_tests` fixtures.
  - Zesu requires canonical Glamsterdam inputs and, with default artifacts, supports only ZisK.
  - Zilkworm remains on the legacy generated-fixture path.
  - Expected public values are now the raw `statelessOutputBytes`, rather than their SHA-256 digest.


crates/benchmark-runner/src/stateless_validator/fixtures.rs
```diff
@@ -337,6 +333,24 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn legacy_fixtures_are_rejected_for_canonical_clients() -> Result<()> {
```
jsign, line 337: (Not for this PR)
I think at some point we might need to create a new workload fixture that is generated from `witness-generator-spec-cli` to also run mainnet/devnet blocks, no? (given now that Reth and Ethrex use canonical ones).

I think like the BPO2 fixtures you created for `ere-guests`. That would prob mean including in `witness-generator-spec-cli` a run of the EEST spec to get also the `statelessOutputBytes`. Is this how you generated them? (See Edit2 below)

I feel this might seem like a good idea to have that generator for ere-guests refresh of cases, or simply completeness given `witness-generator-cli` I think should be soon gone.

Opinions? :smile: 

Edit: other option is make `witness-generator-spec-cli` to generate the minimal same format as an EEST fixture, e.g.:
```
  {
    "rpc-mainnet-25234200": {
      "network": "mainnet",
      "config": {
        "chainid": "1"
      },
      "blocks": [
        {
          "statelessInputBytes": "0x...",
          "statelessOutputBytes": "0x...",
          "blockHeader": {
            "number": "25234200"
          }
        }
      ]
    }
  }
```
Maybe this is an interesting option, since it would mean accepting only a single format both for EEST tests/benchmarks and also mainnet/devnet.

Edit2: Uhm, for BPO2 I wonder how did you generated `statelessOutputBytes` now, since EEST only has stateless executor for Amsterdam. Maybe you actually used Ethrex?

crates/benchmark-runner/src/stateless_validator/fixtures.rs
```diff
@@ -337,6 +333,24 @@ mod tests {
...
+        let fixture_path = dir.path().join("legacy.json");
+        fs::write(&fixture_path, r#"{"stateless_input": {}}"#)?;
+
+        for client in [
+            ExecutionClient::Reth,
+            ExecutionClient::Ethrex,
+            ExecutionClient::Zesu,
+        ] {
+            let err = load_benchmark_fixtures(&fixture_path, dir.path(), client).unwrap_err();
+            assert!(err.to_string().contains("supported only for Zilkworm"));
```
jsign, line 348: (Not for this PR)

I think we could reconsider Zilkworkm until it accepts canonical guest inputs -- I think it creates some friction in still having to accept legacy fixtures.

I think this would also mean we could remove the legacy `witness-generator-cli`/`witness-generator` and rely only in `witness-generator-spec-cli`. 

Does this makes sense?

crates/benchmark-runner/src/stateless_validator/inputs.rs
```diff
@@ -39,37 +30,44 @@ pub(crate) fn stateless_validator_input_from_fixture(
...
                 )
             }
         },
         BenchmarkFixture::Eest(fixture) => match el {
-            ExecutionClient::Reth | ExecutionClient::Ethrex => {
-                raw_eest_input_from_fixture(fixture, EestExpectedOutput::Sha256)
-            }
-            ExecutionClient::Zesu => raw_eest_input_from_fixture(fixture, EestExpectedOutput::Raw),
+            ExecutionClient::Reth | ExecutionClient::Ethrex => raw_eest_input_from_fixture(fixture),
+            ExecutionClient::Zesu => zesu_input_from_fixture(fixture),
```
jsign, line 42: This is just for checking is an Amsterdam block, since I think they don't accept earlier ones. So it is mainly a check, it ends up calling `raw_eest_input_from_fixture(...)`.

crates/benchmark-runner/src/stateless_validator.rs
```diff
@@ -33,15 +33,19 @@ pub struct BlockMetadata {
...
+
 impl ExecutionClient {
-    /// Returns the version string of the execution client (tag or short commit hash),
-    /// extracted from the resolved `Cargo.lock` at build time.
+    /// Returns the version string associated with the selected guest artifact.
     pub const fn version(&self) -> &'static str {
         match self {
-            Self::Reth => env!("RETH_EL_VERSION"),
-            Self::Ethrex => env!("ETHREX_EL_VERSION"),
+            Self::Reth => RETH_EL_VERSION,
```
jsign, line 45: This is quite unfortunate i.e. not using `EL_VERSION`.

The reason is that the Zilkworm crate imports specific version of Reth crate ecosystem that aren't easily solvable. I think this is another reason why removing Zilkworm "as a crate" in this repo is a good idea (apart from waiting until they accept canonical inputs).

crates/witness-generator-spec-cli/src/builder.rs
```diff
@@ -5,22 +5,28 @@ use alloy_eips::eip2718::Decodable2718;
...
 use alloy_rlp::Decodable;
 use anyhow::{Context, ensure};
-use libssz::SszEncode;
-use libssz_types::SszList;
-use stateless_validator_common::new_payload_request::{
-    BlockAccessList, ConsolidationRequest, ConsolidationRequests, DepositRequest, DepositRequests,
-    ExecutionPayloadV4, ExecutionRequests, ExtraData, NewPayloadRequestAmsterdam,
-    Transaction as PayloadTransaction, Transactions, VersionedHashes, Withdrawal,
-    WithdrawalRequest, WithdrawalRequests, Withdrawals,
+use stateless_validator_common::{
```
jsign, line 8: Nuked the `schema.rs` we had here and leverage `stateless_validator_common` structs as we planned.

crates/witness-generator-spec-cli/src/chain_config.rs
```diff
@@ -0,0 +1,39 @@
+//! Amsterdam chain configuration used by generated stateless inputs.
```
jsign, line 1: I think we can remove this in `tests-zkevm@v0.6.1`, or at least simplify the fixed constants that won't be there.

jsign:
@han0110 LMK your thoughts about this -- not sure if you were already updating to v0.6.1 in ere-guest and had some ongoing work for workload too. If that's the case we can close this, and wait for that if was in flux.

crates/benchmark-runner/src/stateless_validator/fixtures.rs
han0110: For BPO2 I generated the `statelessOutputBytes` by using `lighthouse` to compute the `new_payload_request_root` as a reference value (also to make sure that `zkboost` will work) and assume the block is valid, but probably this is not the most reliable approach.

> other option is make witness-generator-spec-cli to generate the minimal same format as an EEST fixture

I generated the BPO2 fixtures of `ere-guests` in the format of the R2 fixtures you made, but I think it'd be nice if all are using the same format.

Given that we are the main user of BPO2 fixtures, perhaps we can just simply fill the `statelessOutputBytes` by `ere-guests` (mainly for benchmark purpose), and once Glamsterdam is online it can be removed and all will be filled with `execution-specs`

crates/benchmark-runner/src/stateless_validator/fixtures.rs
han0110: Make sense to me

han0110 (approved):
I had `zkevm-benchmark-workload` updated to `v0.13.0` but not as complete as this PR (was distracted to updating `ere-guests` to  `v0.6.0`), so I think it'd be good to merge this first!

LGTM, and agree that we should reconsider Zilkworm given that all the other guests support canonical inputs now, the code could be much simpler I believe.

crates/benchmark-runner/src/stateless_validator/fixtures.rs
jsign: Sounds good!

jsign:
@han0110, small note: in the weekend I did [v0.6.1](https://github.com/ethereum/execution-specs/releases/tag/tests-zkevm%40v0.6.1). Nothing changed in spec/tests -- there was a bug in the filling logic, so only the filling changed compared to v0.6.0.

PR #297, opened 2026-07-14
feat: accept only canonical fixtures guest programs

jsign:
Standardise the benchmark runner on canonical EEST `blockchain_tests` fixtures containing `statelessInputBytes` and `statelessOutputBytes`.

This removes the legacy fixture-conversion pipeline and Zilkworm-specific integration, making canonical EEST bytes the single input format for Reth, Ethrex, and Zesu.

## Changes

- Reject legacy fixtures containing a top-level `stateless_input` with a clear migration error.
- Remove Zilkworm execution-client support, dependencies, downloader logic, and input conversion. Whenever Zilkworm is re-published via `ere-guests` we can re-include it pretty easily.
- Remove the legacy `witness-generator` and `witness-generator-cli` crates, Docker image, CI jobs, and related documentation.
- Require an explicit `--input-folder` for execute and prove actions. Just to be more explicit reg flags.
- Use `ere-guests` metadata for the Reth guest version.
- Update documentation and tests for the canonical-only workflow.

The separate `witness-generator-spec-cli` for producing and publishing canonical stateless inputs remains available.

## Breaking changes

- Existing legacy `stateless_input` fixtures must be replaced with canonical EEST fixtures.
- `--execution-client zilkworm` is no longer available.
- Execute and prove commands must now provide `--input-folder`.
- The legacy witness-generator CLI and container image are discontinued.


crates/benchmark-runner/src/stateless_validator/fixtures.rs
```diff
@@ -170,41 +141,28 @@ where
     })
 }
 
-fn fixture_matches_prefixes(fixture: &BenchmarkFixture, prefixes: Option<&[String]>) -> bool {
+fn fixture_matches_prefixes(fixture: &EestStatelessFixture, prefixes: Option<&[String]>) -> bool {
```
jsign, line 144: No more enum with variants since we support a single format now (EEST).

crates/benchmark-runner/src/stateless_validator/fixtures.rs
```diff
@@ -170,41 +141,28 @@ where
...
-        if !matches!(el, ExecutionClient::Zilkworm) {
-            bail!(
-                "{el:?} supports only canonical blockchain_tests fixtures with statelessInputBytes/statelessOutputBytes; legacy stateless_input fixtures are supported only for Zilkworm"
-            );
-        }
-        let fixture = serde_json::from_value(value)
-            .with_context(|| format!("Failed to parse legacy fixture {}", path.display()))?;
-        return Ok(vec![BenchmarkFixture::Legacy(Box::new(fixture))]);
+        bail!(
+            "legacy fixture format with top-level stateless_input is no longer supported; provide an EEST blockchain_tests fixture containing statelessInputBytes and statelessOutputBytes"
```
jsign, line 161: I think we can leave this for a while just in case somebody else is using it -- maybe in some weeks we can clean up further.

crates/benchmark-runner/src/stateless_validator.rs
```diff
@@ -20,31 +19,19 @@ pub enum ExecutionClient {
...
-const RETH_EL_VERSION: &str = "v2.3.0";
 /// Zesu version republished with the ere-guests v0.13.0 artifacts.
 const ZESU_EL_VERSION: &str = "bal-devnet-7-2026-06-24";
 
 impl ExecutionClient {
     /// Returns the version string associated with the selected guest artifact.
     pub const fn version(&self) -> &'static str {
         match self {
-            Self::Reth => RETH_EL_VERSION,
+            Self::Reth => ere_guests_stateless_validator_reth::EL_VERSION,
```
jsign, line 33: @han0110, came back to using this constant -- this pulls a lot of Reth indirect dependencies, but now it is okay since doens't conflict with previous zilkworm crate.

I think whenever Reth and Ethrex are also re-published in ere-guests, we'll have everything as a `::EL_VERSION` but without pulling dependencies.

Thinking out loud here.

crates/ere-hosts/src/cli.rs
```diff
@@ -153,6 +139,23 @@ fn parse_duration(value: &str) -> Result<Duration, String> {
 }
 
 impl Cli {
+    /// Validate arguments whose requirements depend on the selected action.
+    pub fn validate(&self) -> AnyhowResult<()> {
```
jsign, line 143: Just making `--input-folder` required for prove/execute, but not verify. Prefer explicit flags I think.

crates/witness-generator-cli/src/main.rs
jsign, line 1: Man, feels weird to delete this crate... 🫡

jsign, line 1: As said in previous PR, we should prob do some extra work on the `witness-generator-specs-cli` to provide stateless output bytes to make it a bit more completeish and should be okay.

Cargo.toml
```diff
@@ -113,23 +110,13 @@ ere-util-tokio = { git = "https://github.com/eth-act/ere", tag = "v0.13.0" }
...
 # alloy
-alloy-consensus = { version = "=2.0.5", default-features = false }
-alloy-eips = { version = "=2.0.5" }
-alloy-primitives = { version = "=1.6.0", default-features = false }
-alloy-rpc-types-eth = "=2.0.5"
-alloy-genesis = { version = "=2.0.5", default-features = false }
+alloy-consensus = { version = "2.0.5", default-features = false }
+alloy-eips = { version = "2.0.5", default-features = false }
+alloy-primitives = { version = "1.6.0", default-features = false }
 alloy-rlp = { version = "0.3", default-features = false }
```
jsign, line 120: We still use these crates for `witness-generator-specs-cli`. I think super ideally we could think about removing Reth-deps completely, but I don't think there is rush for this.

jsign:
@han0110 doing a temp check here if you can review how this feels.

I spoke with @developeruche and he is going to work on the ELF publishing in https://github.com/eth-act/zilkworm-stateless so we can republish in ere-guests and include it again here in the same pipeline as the others.

crates/benchmark-runner/src/stateless_validator.rs
han0110: Yeah make sense, I'll try to create a crate like `ere-catalog` in `ere-guests` repo but use `build.rs` to extract the version from the `Cargo.toml` and republish registry, so we don't pull the deps of ethrex or reth.

crates/witness-generator-cli/src/main.rs
han0110: 🫡

han0110: Make sense, should we use `execution-specs` to fill `stateless_output_bytes` or we can simply use `ere-guests`?

docs/benchmark-execution-output.md
```diff
@@ -56,10 +56,17 @@ A successful execution metrics file has this shape:
 
 ```json
 {
-  "name": "rpc_block_23743854",
+  "name": "tests_foo.py_test_case_param-block-0-a1b2c3d4",
```
han0110: I found the name format is updated to `"eest__{sanitized}__block{block_index}"` so this seems to be outdated? Not very important to fix I think.

han0110 (approved):
LGTM!

Also I realized that the `tests-benchmarks@vX.Y.Z` doesn't have `statelessInputBytes` and `statelessOutputBytes`, is the idea to upstream the stateless guest to main branch so future release will contain that?

I think thinking now without the `witness-generator-cli` how would we do the benchark with EEST suites.

crates/witness-generator-cli/src/main.rs
jsign: I think for Amsterdam blocks, ideally we use `execution-specs` (maybe re-using-ish the current `./scripts` file that is used for CI to verify the R2 bucket; but maybe there is a more elegant way)

For non-amsterdam blocks (i.e. mainnet) -- I don't have a strong opinion, but as you said maybe we can assume `successful_validation=True` and keep it simple. Could be `ere-guests` too with Ethrex I guess but that might be heavy and not sure would conclude something else than the other idea.

docs/benchmark-execution-output.md
jsign: Good catch! Fixed.

jsign:
> Also I realized that the `tests-benchmarks@vX.Y.Z` doesn't have `statelessInputBytes` and `statelessOutputBytes`, is the idea to upstream the stateless guest to main branch so future release will contain that?
> 
> I think thinking now without the `witness-generator-cli` how would we do the benchark with EEST suites.

@han0110 , Ah, the plan is using Geth with `--evm-bin` to fill the `projects/zkevm-benchmark` branch only for fixtures. I tested this some weeks ago and worked fine.

But now, with a fix I had to do last weekend, it won't work as expected without an extra fix. Also, geth requires a small patch to their `glamsterdam-devnet-x` for t8n.

I'll keep this PR on hold a bit, until I order those pieces and share with you again how I can automate this: maybe with a script here that: i) Pulls `tests-zkevm@vA.B.C` branch, i) Pulls Geth for `glamsterdam-devnet-X` and applies the small patch, iii) Fill the fixtures.

That's mostly the skeleton. I'll work on this later today, and prob make an extra commit to this PR to not leave this gap open. Also interested in you to try it out before merging.

Eventually I can also try doing that script in the EEST branch we have and add it as part of the release process. Just not sure if the CI will be fine with it. But we can go a bit in steps. 

I'll ping later today/tomorrow. LMK if something doesn't make sense.

han0110:
> the plan is using Geth with `--evm-bin` to fill the `projects/zkevm-benchmark` branch only for fixtures

I see, now it makes sense to me!

jsign:
Please see https://github.com/ethereum/execution-specs/issues/3180

I think we can try to fill manually for a bit -- and at some point decide if we want to add a script to automate the steps mentioned there.

prob we can only try when ere-guests have v0.6.2 compatible guest programs.

Hope this makes sense!

PR #298, opened 2026-07-15
fix: update default catalog URL and EEST ref for R2 stateless inputs for glamsterdam-devnet-7

jsign:


jsign:
Oops, see also https://github.com/eth-act/zkevm-benchmark-workload/pull/299

PR #299, opened 2026-07-15
fix: comment out schedule section in EEST R2 Stateless Inputs workflow

jsign:
Pausing the automatic check since realised when trying to do a [first run](https://github.com/eth-act/zkevm-benchmark-workload/actions/runs/29449097312/job/87466899842) something seems to be wrong.

I'll try to check why this fails, and we can reschedule automatically.

### DIFF
--- a/.github/workflows/eest-r2-stateless-inputs.yml
+++ b/.github/workflows/eest-r2-stateless-inputs.yml
@@ -18,8 +18,8 @@ on:
         required: false
         default: "10"
         type: string
-  schedule:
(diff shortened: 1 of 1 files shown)


jsign:
@han0110 , ah, right, we're still in v0.4.1 reg the new SCHEMA_ID and ChainConfig stuff in the `witness-generator-spec-cli` stuff.

I think we can leave this off until we update stuff to v0.6.2.

I'll stop the exporter machine but all is quite ready on that side to have the live R2 bucket :)

No rush anyway.

PR #300, opened 2026-07-17
feat: add opcode_count to eest block metadata

han0110:
After https://github.com/ethereum/execution-specs/pull/3183 is merged, I'd expect future `tests-benchmark@vX.Y.Z` or `tests-zkevm-benchmarks@vX.Y.Z` (or local filled fixtures) will contain `_info.metadata.opcode_count_per_block`. This PR includes that info into `EestBlockMetadata` so the `zkevm-metrics` can be used by `evm-gasfit` alone without having access to the benchmark fixtures.

### DIFF
--- a/crates/benchmark-runner/src/stateless_validator/eest.rs
+++ b/crates/benchmark-runner/src/stateless_validator/eest.rs
@@ -18,6 +18,7 @@ pub(crate) struct EestStatelessFixture {
     pub(crate) chain_id: u64,
     pub(crate) block_number: Option<u64>,
     pub(crate) block_used_gas: Option<u64>,
+    pub(crate) opcode_count: BTreeMap<String, u64>,
     pub(crate) stateless_input_bytes: Vec<u8>,
     pub(crate) stateless_output_bytes: Vec<u8>,
 }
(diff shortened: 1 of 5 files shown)


jsign:
Oh, I see Louis merged it into `devnets/glamsterdam/6-benchmarks` instead of `forks/amsterdam`.

A priori I found that weird, but actually I think can be easier for us since in any upcoming 
[`tests-glamsterdam-devnet@v7.2.x`](https://github.com/ethereum/execution-specs/releases/tag/tests-glamsterdam-devnet%40v7.2.0), then `projects/zkevm-releases` will have it at least.

Thinking a bit if we want maybe to wait to happen naturally; but I could also cherry pick into `projects/zkevm-releases` (would be a bit weird but could unblock this).

han0110:
hmm I see, didn't notice it's PR target.

I think we could probably wait for it to land to `tests-glamsterdam-devnet@v7.2.x`, and if I need to do benchmark before that I can cherry-pick and local fill I think.

This PR should be fine without the `_info.metadata.opcode_count_per_block` field tho (empty if the fixture doesn't contain it).

jsign:
Makes sense, let's wait for next `tests-glamsterdam-devnet@v7.2.x` instead of cherry picking stuff in `projects/zkevm-releases`.

jsign (approved):

jsign:
@han0110 feel free to merge whenever you want if it is fine anyway then, up to you if it simplifies!

PR #301, opened 2026-07-20
Update to tentative ere-guests v0.14

jsign:
_Note: this PR will be kept as a draft until `ere-guests` tags `v0.14` prob when Zesu is supported. ~For now mainly trying to frontrun trying the R2 bucket setup for `glamsterdam-devnet-7`~ (Edit: done, looks like works correctly)_

Update the benchmark workload to the tentative ERE / `ere-guests` v0.14 stack and the `tests-zkevm` v0.6.2 stateless-input format.

- Pin ERE and `ere-guests` to the revisions carrying the Glamsterdam updates, align the Alloy dependencies, and add the temporary `revm`, `alloy-evm`, and Reth compatibility patches required by those revisions.
- Migrate Amsterdam stateless inputs from the generic `0x0001` schema prefix to the fork-qualified `0x1501` prefix. Artifact metadata now derives the schema ID from the generated bytes instead of hard-coding it.
- Extend Gloas input generation to deserialize and encode builder deposit and builder exit requests.
- Use the fork returned by canonical input decoding for Zesu validation and retain raw input forwarding for Reth and Ethrex.
- Stop applying OpenVM-style 32-byte public-output padding to Airbender; OpenVM and ZisK retain their existing normalization.
- Temporarily gate all Zesu runs until a `tests-zkevm` v0.6.2-compatible guest artifact is published, while preserving its routing and compatibility code for re-enablement.
- Update the benchmark and stateless-input publication documentation for the new schema, EEST version, and Zesu availability.

## Compatibility and follow-ups

- Newly generated Amsterdam inputs begin with `0x1501`; fixture bundles and artifacts using the previous `0x0001` format need to be regenerated with the v0.6.2-compatible tooling.
- The ERE / `ere-guests` revision pins and `[patch.crates-io]` entries are temporary and should be replaced or removed once the corresponding releases are available. Prob we might be able to remove them when Reth publishes ELFs and we can stop having to apply patches that have to be downstreamed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,20 +103,20 @@ zkevm-metrics = { path = "crates/metrics" }
 benchmark-runner = { path = "crates/benchmark-runner" }
 
 # ere
-ere-dockerized = { git = "https://github.com/eth-act/ere", tag = "v0.13.0" }
-ere-cluster-client-zisk = { git = "https://github.com/eth-act/ere", tag = "v0.13.0" }
-ere-util-tokio = { git = "https://github.com/eth-act/ere", tag = "v0.13.0" }
+ere-dockerized = { git = "https://github.com/eth-act/ere", rev = "58ca85beaee2fa8acd31dbf33b90bb765aac9010" }
+ere-cluster-client-zisk = { git = "https://github.com/eth-act/ere", rev = "58ca85beaee2fa8acd31dbf33b90bb765aac9010" }
+ere-util-tokio = { git = "https://github.com/eth-act/ere", rev = "58ca85beaee2fa8acd31dbf33b90bb765aac9010" }
 
 # ere-guests
-stateless-validator-common = { git = "https://github.com/eth-act/ere-guests", tag = "v0.13.0", package = "stateless-validator-common" }
-ere-guests-stateless-validator-ethrex = { git = "https://github.com/eth-act/ere-guests", tag = "v0.13.0", package = "stateless-validator-ethrex" }
-ere-guests-stateless-validator-reth = { git = "https://github.com/eth-act/ere-guests", tag = "v0.13.0", package = "stateless-validator-reth" }
-ere-guests-downloader = { git = "https://github.com/eth-act/ere-guests", tag = "v0.13.0", package = "downloader" }
+stateless-validator-common = { git = "https://github.com/eth-act/ere-guests", rev = "b2ed4d88a30174b1eb5d75c55b63deab8516e8bd", package = "stateless-validator-common" }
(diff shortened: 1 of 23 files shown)


jsign:
Looks like all is working well in the [R2 bucket validated by eest stateless guest](https://github.com/eth-act/zkevm-benchmark-workload/actions/runs/29766777973/job/88434835507#step:7:82)! ([R2 bucket](https://pub-df22334654034ebab51bc096137a59d8.r2.dev/devnets/glamsterdam-devnet-7/index.html)) (Edit: the bucket might be broken when this is read, see next sentence)

Edit: now getting a bit more ambitious to close the loop in simplification and benchmarking, see https://github.com/eth-act/zkevm-benchmark-workload/pull/302 (which targets this branch).

crates/benchmark-runner/src/guest_programs.rs
```diff
@@ -97,9 +97,7 @@ fn normalize_expected_public_values(
     zkvm_kind: zkVMKind,
     mut expected_public_values: Vec<u8>,
 ) -> Vec<u8> {
-    if matches!(zkvm_kind, zkVMKind::Airbender | zkVMKind::OpenVM)
-        && expected_public_values.len() < 32
-    {
+    if matches!(zkvm_kind, zkVMKind::OpenVM) && expected_public_values.len() < 32 {
         expected_public_values.resize(32, 0);
     }
```
han0110: This is no longer necessary btw, we configure OpenVM to have 256 bytes output and 3 zkVMs now should be good to support `stateless_output_bytes`

Cargo.toml
```diff
@@ -146,3 +146,24 @@ tar = "0.4"
 toml = "0.8"
 hex = "0.4"
 zstd = "0.13"
+
+# Temporary patches required by the Reth revision used in ere-guests.
+# Remove these once Glamsterdam devnet support is available in released crates.
+[patch.crates-io]
```
han0110: Perhaps this is not needed since we only use the `EL_VERSION` from `stateless-validator-reth`? I'll create a crate soon to just exports these version values so we don't need to pull in those deps btw.

crates/benchmark-runner/src/guest_programs.rs
jsign: Makes sense, fixed in https://github.com/eth-act/zkevm-benchmark-workload/pull/301/commits/2fde2482812be843e929804b9a54aedf4ea753ce
Did something similar to ere-guests output matcher.

Cargo.toml
jsign: Removed in https://github.com/eth-act/zkevm-benchmark-workload/pull/301/changes/dc6c58650c985204b8f52d3010aef32ee6166a0a

jsign:
@han0110 prefered to not leave too many things in the air -- whenever Zesu is added to ere-guests and we tag, we can re-enable it here. (And prob start aligning releases in workload and ere-guests, I think would be useful)

PR #302, opened 2026-07-20
Use EEST fixtures as the unified witness-generator output format

jsign:
_Note: this is targeting https://github.com/eth-act/zkevm-benchmark-workload/pull/301 branch, just to keep things clean from a merging perspective. ~Still pending to try it out in `glamsterdam-devnet-7` and R2 bucket with new format.~_

This PR makes EEST blockchain-test fixtures the unified output format of `witness-generator-spec-cli`.

The main goal is consistency across the stateless-validation toolchain:

- EL implementations already use EEST fixtures as the common interoperability format.
- `ere-hosts` already discovers and consumes EEST fixtures directly.
- The witness generator can now produce the same format from live CL/EL networks, avoiding a separate generator-specific artifact schema and conversion step.

This means the complete workflow is now:

```text
collect live blocks → EEST fixtures → export batch → extract → ere-hosts
```

## Expected stateless output

The generator now produces both `statelessInputBytes` and the expected `statelessOutputBytes`.

The expected output does not require executing the block. After constructing the `StatelessInput`, the generator:

1. validates the generated chain configuration against the payload;
2. calculates the SSZ hash-tree-root of `new_payload_request` using SHA-256;
3. constructs a successful `StatelessValidationResult` containing that root and the input chain configuration;
4. serializes the result as plain SSZ, without an input schema prefix.

This provides the complete expected public output required by `ere-hosts` while leaving full execution validation to the stateless guest.

## Unified EEST fixture format

Both one-shot generation and live collection use the same fixture serializer.

Each fixture contains:

- `network: "Amsterdam"`;
- hexadecimal `config.chainid`;
- one block with `statelessInputBytes` and `statelessOutputBytes`;
- block number and gas used;
- block hash, slot, generator provenance, schema version, and collection metadata where applicable.

`generate` now always outputs a benchmark-ready EEST JSON fixture. The old input-only `hex` and `raw` output modes and the `--format` option were removed.

Live collection stores each fixture as an individually compressed `.json.zst` file. The decompressed content is a complete EEST fixture rather than a custom artifact object.

## Benchmark-ready exports

Exported `.tar.zst` batches now use this layout:

```text
blockchain_tests/<chunk>/<block-hash>.json
.meta/manifest.json
```

Fixtures inside the archive are uncompressed JSON; compression is provided by the outer archive.

After extraction, the archive root can be passed directly to:

```bash
ere-hosts ... stateless-validator --input-folder <extracted-root>
```

No conversion or fixture preparation step is required.

The batch manifest records fixture paths, fixture checksums and lengths, block metadata, and input length. It lives under `.meta/` so `ere-hosts` fixture discovery does not interpret it as a blockchain test.

### DIFF
--- a/.github/workflows/eest-r2-stateless-inputs.yml
+++ b/.github/workflows/eest-r2-stateless-inputs.yml
@@ -1,4 +1,4 @@
-name: EEST R2 Stateless Inputs
+name: EEST R2 Stateless Fixtures
 
 on:
   workflow_dispatch:
@@ -62,7 +62,7 @@ jobs:
         id: eest
         run: echo "commit=$(git -C execution-specs rev-parse HEAD)" >> "$GITHUB_OUTPUT"
 
-      - name: Validate R2 stateless inputs with EEST
+      - name: Validate R2 stateless fixtures with EEST
         run: |
           mkdir -p "$SUMMARY_DIR"
           uv run --project execution-specs --with zstandard \
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6755,6 +6755,7 @@ dependencies = [
  "alloy-primitives",
  "alloy-rlp",
  "anyhow",
+ "benchmark-runner",
  "clap",
  "humantime",
  "reqwest",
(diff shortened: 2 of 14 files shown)


jsign:
Trying out in a fresh R2 bucket with this new format... posting here later updates.

Update 1: looks it it works, see [here R2 bucket](https://pub-df22334654034ebab51bc096137a59d8.r2.dev/devnets/glamsterdam-devnet-7/index.html) with for `glamsterdam-devnet-7` and this "new" format ("new" == use EEST fixture format).
Update 2: validation in CI against EEST stateless guest is [green](https://github.com/eth-act/zkevm-benchmark-workload/actions/runs/29774945159/job/88461972730). Although I see takes some time since this devnet now allows some 300M gas blocks ha

PR #305, opened 2026-07-22
Update ere-guests and use catalog

jsign:
Update `ere-guests` and now use change introduced in https://github.com/eth-act/ere-guests/pull/66


han0110 (approved):
LGTM

PR #307, opened 2026-07-28
feat: load only last block of eest fixture, and add target_opcode

han0110:
- In `load_eest_benchmark_fixtures` load only last block, to avoid loading 256 setup blocks from each of 5 `tests/benchmark/compute/instruction/test_block_context.py::test_blockhash[...]` fixtures, that creates unuseful workload to benchmark runner.
- Add `target_opcode` to `metadata` of zkevm-metrics if present in EEST fixture, that should've added in #300 (for `evm-gasfit` integration)

### DIFF
--- a/crates/benchmark-runner/src/stateless_validator/eest.rs
+++ b/crates/benchmark-runner/src/stateless_validator/eest.rs
@@ -18,7 +18,8 @@ pub(crate) struct EestStatelessFixture {
     pub(crate) chain_id: u64,
     pub(crate) block_number: Option<u64>,
     pub(crate) block_used_gas: Option<u64>,
-    pub(crate) opcode_count: BTreeMap<String, u64>,
+    pub(crate) opcode_count: Option<BTreeMap<String, u64>>,
+    pub(crate) target_opcode: Option<String>,
     pub(crate) stateless_input_bytes: Vec<u8>,
     pub(crate) stateless_output_bytes: Vec<u8>,
 }
@@ -48,7 +49,9 @@ struct EestInfo {
 #[derive(Debug, Default, Deserialize)]
 struct EestMetadata {
     #[serde(default)]
-    opcode_count_per_block: Vec<BTreeMap<String, u64>>,
+    opcode_count_per_block: Option<Vec<BTreeMap<String, u64>>>,
+    #[serde(default)]
+    target_opcode: Option<String>,
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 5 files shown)


crates/benchmark-runner/src/stateless_validator/eest.rs
```diff
@@ -449,31 +464,41 @@ mod tests {
...
+            "tests/foo.py::test_same[empty_input]": {
+                "network": "Amsterdam",
+                "config": {"chainid": "0x01"},
+                "blocks": [
+                    {
+                        "statelessInputBytes": "0x",
+                        "statelessOutputBytes": "0xcc"
+                    }
+                ]
+            },
```
han0110, line 476: Add to test the empty input skipping branch, which was in the `tests/foo.py::test_same[name/a].blocks[2]`.

crates/benchmark-runner/src/stateless_validator/eest.rs
```diff
@@ -18,7 +18,8 @@ pub(crate) struct EestStatelessFixture {
     pub(crate) chain_id: u64,
     pub(crate) block_number: Option<u64>,
     pub(crate) block_used_gas: Option<u64>,
-    pub(crate) opcode_count: BTreeMap<String, u64>,
+    pub(crate) opcode_count: Option<BTreeMap<String, u64>>,
+    pub(crate) target_opcode: Option<String>,
```
jsign, line 22: Ah yes, this is helpful.

The only thing I remember is that for precompiles benchmarks, EEST decided to put `STATICCALL` here, whichi is technically correct since precompiles are called with `STATICCALL`. 

I think I raised this point to Louis once and I think they were considering to override it to have meaningful tags for precompiles. I remember noting this when also thinking of `evm-gasfit` integration.

I don't remember if this happened or not -- prob if happened `opcode` naming might not be ideal. But not a big deal if that's the case.

In any case, I guess `evm-gasfit` can always do some extra parsing and "fix" this case.  Just sharing this that I remembered now.

docs/benchmark-execution-inputs.md
```diff
@@ -70,10 +71,12 @@ Rules:
...
+- Only the last block of a test case is loaded, because EEST benchmark tests place the worst case block last and use any preceding blocks for setup.
 - A block without `statelessInputBytes`, or with empty `statelessInputBytes`, is skipped.
 - A block with non-empty `statelessInputBytes` must also contain `statelessOutputBytes`.
 - Both byte fields are hexadecimal strings with an optional `0x` prefix and an even number of hexadecimal digits after that prefix.
-- `_info.metadata.opcode_count_per_block` holds one opcode-count map per block in block order, so entry N describes `blocks[N]`. A present array whose length differs from the block count is rejected, and an absent array leaves each block's opcode count empty.
+- `_info.metadata.opcode_count_per_block` holds one opcode-count map per block in block order, so entry N describes `blocks[N]` and the last entry describes the loaded block. A present array whose length differs from the block count is rejected.
+- `_info.metadata.target_opcode` names the opcode a benchmark stresses. Benchmarks that stress no single opcode omit it.
```
jsign, line 79: Uhm, maybe for precompiles now then is not STATICCALL then?

jsign (approved):
LGTM! 
Just shared something I raised before but mainly as question since not sure how evolved with time.

crates/benchmark-runner/src/stateless_validator/eest.rs
han0110: Ahh right, didn't think about that, the local one I generated now filled with the precompile name, so probably they already change it (seem to be [this PR](https://github.com/ethereum/execution-specs/pull/2591)?). But the `opcode_count_per_block` still shows the `STATICCALL` count tho, I guess we'd need to trust that number are all from precompile calls.

Ah just remember in [`eest.py`](https://github.com/misilva73/evm-gasfit/blob/main/src/evm_gasfit/adapter/eest.py) it already does the extra parsing, will wrap up the `evm-gasfit` integration tomorrow and figure all these out.

jsign: Yep, I remember this being tricky. Honestly, feels to me this is a generic situation rather than a zkEVM one.

Maybe we can assume STATICCALL count == opcode target (or should say `opcode/precompile target`). Although it sounds dangerous-ish.

Prob worth raising this to Maria and/or Louis; not sure it should be blocking for this PR.

han0110: yeah I'd just assume that for now

PR #311, opened 2026-09-10
Warn on mismatched EEST opcode count metadata

jsign:
The EEST loader previously failed when `opcode_count_per_block` had a different number of entries than `blocks`, even when the guest input and output were valid. This change logs a warning, omits opcode counts for the affected test case and continues loading the fixture.

Updates the input and output documentation and adds test coverage for empty, shorter, matching and longer opcode count arrays.

I discovered this while running a [new Hive UI dashboard](https://github.com/eth-act/eest-execution-witness-dashboard/actions/runs/34525355957/job/103032838399#step:10:40507). Quite honestly, I'm not sure this makes sense from the EEST perspective so for now not making this situation a blocker, but I'll dig into EEST why this happen and if makes sense.

### DIFF
--- a/crates/benchmark-runner/src/stateless_validator/eest.rs
+++ b/crates/benchmark-runner/src/stateless_validator/eest.rs
@@ -4,7 +4,7 @@ use std::{
     collections::{BTreeMap, HashSet},
     path::Path,
 };
-use tracing::info;
+use tracing::{info, warn};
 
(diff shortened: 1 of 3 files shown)


jsign:
Took a quick look and I think indeed there might be an eest bug in their default branch.
I'll see to make a minimal reproducer tomorrow to double check.

PR #312, opened 2026-09-14
feat: witgen cli auth/zstd/continuous

han0110:
- Add the following fields in config
  - `el_headers` and `cl_headers`  to support auth
  - `continuous` and `max_concurrency` to support collect continuous blocks since head, with concurrentcy if the witness is serving is slower than slot time.
  - `zstd_window_log` to suppoer long distance window, for mainnet using 2 GiB window helps to increase the compression ratio from ~2.6 to ~9.5. De/compression uses more RAM (~the widnow size), compression time is much slower, but the decompression time stays similar (a bit faster actually).
- WIth all these additional support we can collect mainnet blocks and use github release as the r2 bucket to publish the fixtuers for testing/benchmarking (need some CL api tweaks to support pre-amsterdam payload fetching), an example of 10000 mainnet blocks can be found [here](https://github.com/han0110/benchmarkoor/releases/tag/mainnet-stateless-inputs-25950000-25959999)


crates/witness-generator-spec-cli/src/collector.rs
```diff
@@ -77,6 +83,103 @@ async fn collect_head_once(
     Ok(persisted)
 }
 
+/// Collects every block from the chain tip at start-up onward, in block order, so no height is
+/// skipped while the collector runs.
+async fn collect_continuously(
```
jsign: Uhm, I would say we could probably remove the polling style and just leave "continously" as the only option since I don't see any benefits of polling compared to this new feature, plus no downsides of removing it. (we could simply use `max_concurrency = 1` if want a lightweight setup).

Maybe that can simplify some code further, unless maybe i'm missing something. WDYT?

jsign (approved):
LGTM! left a comment just for your consideration, no need to do in this pr if makes sense.

crates/witness-generator-spec-cli/src/collector.rs
han0110: > since I don't see any benefits of polling compared to this new feature

Actually there is one downside of requiring continuous, so if we set `max_concurrency = 1` but the EL can't serve witness fast enough, the fetching will fall behind more and more, upon certain distance the EL can no longer serve due to timeout and the cli will stuck. And the origianl fetching head approach will not have this issue.

On mainnet I have to set `max_concurrency = 2` to download 10000 continous blocks (probably 3 or 4 will be safer), on devnet I haven't tried yet tho. Do you think this is acceptable?

jsign: I think should be fine. I think there is an expectation of batches not having gaps, thus not missed.

I think should be okay to let the operator run with some reasonable concurrency setup to try to catch-up fast using prob many available cores.

Maybe my concurrency 1 example wasn't ideal since if you set up that for any intentional reason, prob your hw setup is too low for running the daemon.

han0110: I see, then I'd set the default `max_concurrency = 2`, will update it soon. btw I'm using pandsops' reth node.

jsign: Ah nice, I supposed since you were adding auth headers 😁

jsign: For devnet, I use the witness collector machine with ethrex and prysm joining locally, plus the systemd config. :)

Porb might be good to document this, although I think setup from scratch is quite easy.

I should prob think if pandaops have devnet ones that we could use, and maybe run the collector in CI. Although feels maybe a bit of a stretch for scheduled CI 😂 but always feels nicer to avoid VMs.

han0110: > For devnet, I use the witness collector machine with ethrex and prysm joining locally, plus the systemd config.

I see I see, saw the document but never tried to spin that up since you already provide the bucket :)

---

Did a cleanup in https://github.com/eth-act/zkevm-benchmark-workload/pull/312/commits/46dca5d3b651d31adcfe814d7a82d7fe277a44ea, and set the default `max_concurrency = 2` (to not DoS the EL too much I guess)

